PR #10, opened 2019-01-08
corrects formating to g_message of file size

Raphexion:
Problem is that the g_message call at hawkbit-client.c:618
is passing the file size as %ld in format. Must be %lld due
to the data type holding the file size. If this is not changed
the application will segfault in a strlen call in extension of
the g-message call.

Also the g_strdup_printf call at hawkbit-client.c:624 should
have the %ld changed to %lld.

Keeping the %ld entries also produces compiler warnings in GCC 8.2.0

Signed-off-by: Niklas Johansson <niklas.johansson@prevas.dk>

### DIFF
--- a/src/hawkbit-client.c
+++ b/src/hawkbit-client.c
@@ -615,13 +615,13 @@ gboolean process_deployment(JsonNode *req_root, GError **error)
                 goto proc_error;
         }
 
-        g_message("New software ready for download. (Name: %s, Version: %s, Size: %ld, URL: %s)",
+        g_message("New software ready for download. (Name: %s, Version: %s, Size: %" G_GINT64_FORMAT ", URL: %s)",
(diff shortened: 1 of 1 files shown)


src/hawkbit-client.c
```diff
@@ -615,13 +615,13 @@ gboolean process_deployment(JsonNode *req_root, GError **error)
                 goto proc_error;
         }
 
-        g_message("New software ready for download. (Name: %s, Version: %s, Size: %ld, URL: %s)",
```
ejoerns: What about simply using the G_GINT64_FORMAT define here?

    g_message("New software ready for download. (Name: %s, Version: %s, Size: %"G_GINT64_FORMAT ", URL: %s)"

sknsean:
LGTM

PR #11, opened 2019-01-09
rest_request: clean up curl on malloc failure

nsjodk:
Signed-off-by: Niklas Johansson <niklas.johansson@prevas.dk>


sknsean:
LGTM

PR #12, opened 2019-01-11
add meson build option

sknsean:


sknsean:
@hundeboll okay?

nsjodk:
It looks good to me. However I cannot build it on Ubuntu 18.04.1 LTS.
As we keep the original cmake build I am not against merging this.
The only risk I see is that we need to keep two build systems up-to-date.
(For example if someone adds a c-file, it needs to get added to both).

ejoerns:
I would also like to avoid maintaining two build systems at once in the long run.

Thus, when thinking about taking this PR it would mean IMHO: Keep both CMake and meson in parallel for a limited period of time (to give integrators time to migrate) and then drop CMake in favor of meson. The period should last at least for one release.

I see good arguments for meson, as it is way more modern and intuitive than CMake is (at least for me, but I assume I am not alone).

The questions would be: Are there good reasons for keeping CMake instead?
And, is meson better enough to makes it worth putting effort in moving from one to the other?

Opinions are welcome!

ejoerns:
No more progress or comments here so far, I will close this in favor of #113 which partly builds on top of this.

PR #14, opened 2019-12-04
make internal functions static

ejoerns:
This is a collection of patches making functions only internally used static.

This prepares for allowing more picky compiler checks and gain more robust and optimized code.


prevas-lkmi:
LGTM, thanks @ejoerns

PR #15, opened 2019-12-04
Add support for uncrustify code beautifier

ejoerns:
This gives the code a consistent coding style.

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -14,7 +14,7 @@ static gboolean output_to_systemd = FALSE;
  *
  * @param[in] level Log level that should be returned as string.
  */
-static const gchar *log_level_to_string (GLogLevelFlags level)
+static const gchar *log_level_to_string(GLogLevelFlags level)
 {
         switch (level) {
         case G_LOG_LEVEL_ERROR:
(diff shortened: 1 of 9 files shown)


sknsean:
LGTM :+1:

src/log.c
```diff
@@ -14,23 +14,23 @@ static gboolean output_to_systemd = FALSE;
...
+                        return "WARNING";
+                case G_LOG_LEVEL_MESSAGE:
+                        return "MESSAGE";
+                case G_LOG_LEVEL_INFO:
+                        return "INFO";
+                case G_LOG_LEVEL_DEBUG:
+                        return "DEBUG";
+                default:
+                        return "UNKNOWN";
         }
```
Raphexion: Do we want this style? Compare to the linux style guide where the cases are aligned with the switch.
https://www.kernel.org/doc/html/v4.10/process/coding-style.html

ejoerns: As my first intention was to align coding style instead of changing it, I could simply push a fix for the switch if you like.

Raphexion: I really like the changes you made. This was the only issue I wanted to "discuss" before we approve. I think that @prevas-lkmi should decide, because he wrote 99.9% the code.

But I summary, great PR :sunny:

Raphexion: @ejoerns Sorry, I didn't answer your suggestion. I prefer the original style, so if you agree, you can push a fix where the style goes back.

ejoerns: Done ;)

Raphexion:
LGTM :+1:

PR #16, opened 2020-01-08
Add minimal shell test suite and allow testing against docker hawkbit instance

ejoerns:
This series consists of mutliple pre-work and adaptions required for running a minimum set of runtime tests for rauc-hawkbit-updater.

* Add sharness shell test suite
* Prepare code for returning error codes in ase of failures
* spawn hawkbit using docker from travis test
* simple tests for token-based authentification with hawkbit

### DIFF
--- a/src/rauc-hawkbit-updater.c
+++ b/src/rauc-hawkbit-updater.c
@@ -151,7 +151,7 @@ int main(int argc, char **argv)
 
         setup_logging(PROGRAM, log_level, opt_output_systemd);
         hawkbit_init(config, on_new_software_ready_cb);
-        hawkbit_start_service_sync();
+        exit_code = hawkbit_start_service_sync();
 
         config_file_free(config);
 out:
--- a/src/rauc-hawkbit-updater.c
+++ b/src/rauc-hawkbit-updater.c
@@ -118,6 +118,12 @@ int main(int argc, char **argv)
                 goto out;
         }
 
+        if (config_file == NULL) {
+                g_printerr("No configuration file given\n");
+                exit_code = 2;
+                goto out;
+        }
+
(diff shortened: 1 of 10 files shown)


src/rauc-hawkbit-updater.c
```diff
@@ -118,6 +118,12 @@ int main(int argc, char **argv)
                 goto out;
         }
 
+        if (config_file == NULL) {
+                g_printerr("No configuration file given\n");
+                exit_code = 3;
```
Raphexion: Would it be better to change the error_code to something different than 3? Three is already used below (No such configuration file).

Raphexion:
I think it is a good pull request. Thank you for the well written commit messages.

LGTM. @prevas-lkmi has the last say.

src/rauc-hawkbit-updater.c
ejoerns: I also thought about that, but it was not clear to me what the error code says and I interpreted it as 'configuration file issue'. We could also return 4 here, but maybe we need a better definition of error return codes, anyway. I do not have a real strong opinion on this, yet. What would you prefer?

Raphexion: I don't have a strong opinion. Imagine that we change it to 4. Could we then create an extra test case? I am thinking that we can have two different errors:
1. No configure file provided (null).
2. Config file provided but it does not exist.

ejoerns: I've now changed the exit code and also added three more tests covering the cases a) invalid argument, b) missing config argument, c) path to non-existing config

Raphexion: Great!

PR #20, opened 2020-02-25
Gateway token support

serg-z:
Authentication with gateway token: https://www.eclipse.org/hawkbit/concepts/authentication/

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,3 +99,25 @@ Application Options:
   -r, --run-once           Check and install new software and exit
   -s, --output-systemd     Enable output to systemd
 ```
+
+
+Authentication
+--------------
+
+As described on the [hawkBit Authentication page](https://www.eclipse.org/hawkbit/concepts/authentication/)
+in the "DDI API Authentication Modes" section, a device can be authenticated
+with a security token. A security token can be either a "Target" token or a
+"Gateway" token. The "Target" security token is specific to a single target
+defined in hawkBit. In the RAUC hawkBit updater's configuration file it's
+referred to as "auth_token".
+
+Targets can also be connected through a gateway which manages the targets
+directly and as a result these targets are indirectly connected to the hawkBit
+update server. The "Gateway" token is used to authenticate this gateway and
+allow it to manage all the targets under its tenant. With RAUC hawkBit updater
+such token can be used to authenticate all targets on the server. I.e. same
+gateway token can be used in a configuration file replicated on many targets.
+In the RAUC hawkBit updater's configuration file it's called "gateway_token".
+Although gateway token is very handy during development or testing, it's
+recommended to use this token with care because it can be used to
+authenticate any device.
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ Setup target (device) configuration file:
   tenant_id                 = DEFAULT
   target_name               = test-target
   auth_token                = bhVahL1Il1shie2aj2poojeChee6ahShu
+  #gateway_token            = bhVahL1Il1shie2aj2poojeChee6ahShu
   bundle_download_location  = /tmp/bundle.raucb
   retry_wait                = 60 ; 1 min. retry wait
   connect_timeout           = 20 ; 20 secs. connection timeout
--- a/config.conf.example
+++ b/config.conf.example
@@ -4,7 +4,9 @@ ssl                       = false                               # true = HTTPS, 
 ssl_verify                = false                               # validate ssl certificate (only use if ssl is true)
 tenant_id                 = DEFAULT                             # Tenant id
 target_name               = test-target                         # Target name (controller id),
-auth_token                = cb115a721af28f781b493fa467819ef5    # Security token.
+auth_token                = cb115a721af28f781b493fa467819ef5    # Target security token.
+# Or gateway_token can be used instead of auth_token
+#gateway_token             = cb115a721af28f781b493fa467819ef5    # Gateway security token.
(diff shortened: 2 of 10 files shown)


Raphexion:
@serg-z Thank you for contributing :+1: 

Do you mind adding something in the README.md about the difference between an auth token and a gateway token? A minimal guide. For example, that they cannot be used at the same time. When to use an auth token, when to use the gateway token. I think that would be very appreciated for new users.

serg-z:
> @serg-z Thank you for contributing 
> 
> Do you mind adding something in the README.md about the difference between an auth token and a gateway token? A minimal guide. For example, that they cannot be used at the same time. When to use an auth token, when to use the gateway token. I think that would be very appreciated for new users.

Added new "Authentication" section with info about tokens.

Raphexion:
@serg-z Thank you for the text - perfect!

Just a question, do you think it could be a problem that we already use 4 for another error:
https://github.com/rauc/rauc-hawkbit-updater/blob/84da0e26e5adff7000b60fe22e821570108286a4/src/rauc-hawkbit-updater.c#L160

ejoerns:
> @serg-z Thank you for the text - perfect!
> 
> Just a question, do you think it could be a problem that we already use 4 for another error:
> https://github.com/rauc/rauc-hawkbit-updater/blob/84da0e26e5adff7000b60fe22e821570108286a4/src/rauc-hawkbit-updater.c#L160

I'd say it is time to start using the GError definitions correctl. I.e. defining error domains and error codes. 

The error code also does not necessarily have something to do with the actual return value of the program, it just has a meaning inside its specific domain and we might map them later for gaining the actual return value.

Defining error domains will probably look like this here ('rhu' used for rauc-hawkbit-updater, but it is not mandatory to have such a prefix.).

``` c
#define RHU_CONFIG_ERROR rhu_config_error_quark ()

G_DEFINE_QUARK (rhu-config-error-quark, rhu_config_error)
```

And then you define errors like

``` c
#define RHU_CONFIG_ERROR_FOOBAR    3
#define RHU_CONFIG_ERROR_NO_TOKEN  4
#define RHU_CONFIG_ERROR_SANITY    5
```

Raphexion:
> > @serg-z Thank you for the text - perfect!
> > Just a question, do you think it could be a problem that we already use 4 for another error:
> > https://github.com/rauc/rauc-hawkbit-updater/blob/84da0e26e5adff7000b60fe22e821570108286a4/src/rauc-hawkbit-updater.c#L160
> 
> I'd say it is time to start using the GError definitions correctl. I.e. defining error domains and error codes.
> 
> The error code also does not necessarily have something to do with the actual return value of the program, it just has a meaning inside its specific domain and we might map them later for gaining the actual return value.
> 
> Defining error domains will probably look like this here ('rhu' used for rauc-hawkbit-updater, but it is not mandatory to have such a prefix.).
> 
> ```c
> #define RHU_CONFIG_ERROR rhu_config_error_quark ()
> 
> G_DEFINE_QUARK (rhu-config-error-quark, rhu_config_error)
> ```
> 
> And then you define errors like
> 
> ```c
> #define RHU_CONFIG_ERROR_FOOBAR    3
> #define RHU_CONFIG_ERROR_NO_TOKEN  4
> #define RHU_CONFIG_ERROR_SANITY    5
> ```

It sounds really good to me. Isn't the most appropriate way forward to accept this PR in (after you have reviewed it of course).

Then we implement the improved error codes, like you suggest, in a another PR. That is, no need to think about the "4" right now.

ejoerns:
> It sounds really good to me. Isn't the most appropriate way forward to accept this PR in (after you have reviewed it of course).
> 
> Then we implement the improved error codes, like you suggest, in a another PR. That is, no need to think about the "4" right now.

Yes, the PR at does everything that looks a bit hackish consistent to that what was already there, thus these are issues to fix in a different context.

README.md
```diff
@@ -98,3 +99,23 @@ Application Options:
...
+
+Authentication
+--------------
+
+As described on the [hawkBit Authentication page](https://www.eclipse.org/hawkbit/concepts/authentication/)
+in the "DDI API Authentication Modes" section, a device can be authenticated
+with a security token. A security token can be either a "Target" token or a
+"Gateway" token. The "Target" security token is specific to a single target
+defined in hawkBit. In the RAUC hawkBit updater's configuration file it's
+referred to as "auth_token". The "Gateway" token allows all targets on the
```
ejoerns: Just noted that this is misleading. It is meant to be used for a *Gateway* that handles (and possibly authenticates) multiple targets, it is not primarily meant for directly authenticating targets!

serg-z: > Just noted that this is misleading. It is meant to be used for a _Gateway_ that handles (and possibly authenticates) multiple targets, it is not primarily meant for directly authenticating targets!

Reworded to match the hawkBit documentation.

ejoerns: @serg-z Thank you for adapting the documentation!

serg-z:
Hi @ejoerns and @Raphexion ,

When are you planning to merge this? I'm asking because I have PTXdist package files for rauc-hawkbit-updater that I would like to contribute and I would like to use upstream master with this feature in it before sending out patches.

Thanks,
Sergey

ejoerns:
@serg-z From my point of view, this looks fine and could be merged. I'd also appreciate a PTXdist package. @Raphexion what's your opinion?

Raphexion:
@serg-z @ejoerns It looks good to me. (I don't have the rights to merge - could you merge it @ejoerns ).

PR #22, opened 2020-04-02
src/hawkbit-updater: add GError handling for get_available_space

ejoerns:
Collect and forward the actual error in `get_available_space()`.
Otherwise an invalid download path for example will result in an error
like:

> CRITICAL: Error: Not enough free space. File size: 5661. Free space: -1

With this fix, the same situation will tell:

> CRITICAL: Error: Failed to calculate free space: No such file or directory


prevas-lkmi:
LGTM

PR #24, opened 2020-04-02
Avoid potential NULL pointer dereference

Raphexion:
It is not guaranteed that error is assigned when printing error messages.

### DIFF
--- a/src/config-file.c
+++ b/src/config-file.c
@@ -61,7 +61,7 @@ static gboolean get_key_string(GKeyFile *key_file, const gchar* group, const gch
 static gboolean get_key_bool(GKeyFile *key_file, const gchar* group, const gchar* key, gboolean* value, const gboolean default_value, GError **error)
 {
         g_autofree gchar *val = NULL;
-        val = g_key_file_get_string(key_file, group, key, NULL);
+        val = g_key_file_get_string(key_file, group, key, error);
         if (val == NULL) {
                 *value = default_value;
                 return TRUE;
@@ -81,7 +81,7 @@ static gboolean get_key_bool(GKeyFile *key_file, const gchar* group, const gchar
 
 static gboolean get_key_int(GKeyFile *key_file, const gchar* group, const gchar* key, gint* value, const gint default_value, GError **error)
 {
-        gint val = g_key_file_get_integer(key_file, group, key, NULL);
+        gint val = g_key_file_get_integer(key_file, group, key, error);
         if (val == 0) {
                 *value = default_value;
                 return TRUE;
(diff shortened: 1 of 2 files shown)


src/config-file.c
```diff
@@ -61,7 +61,7 @@ static gboolean get_key_string(GKeyFile *key_file, const gchar* group, const gch
 static gboolean get_key_bool(GKeyFile *key_file, const gchar* group, const gchar* key, gboolean* value, const gboolean default_value, GError **error)
 {
         g_autofree gchar *val = NULL;
-        val = g_key_file_get_string(key_file, group, key, NULL);
+        val = g_key_file_get_string(key_file, group, key, error);
```
ejoerns, line 64: This won't help. It sets error but the line below ensures that the function unconditionally returns `TRUE`, anyway.
Thus there is no chance that this error would be handled.

src/rauc-hawkbit-updater.c
```diff
@@ -125,8 +125,12 @@ int main(int argc, char **argv)
         context = g_option_context_new("");
         g_option_context_add_main_entries(context, entries, NULL);
         if (!g_option_context_parse_strv(context, &args, &error)) {
-                g_printerr("option parsing failed: %s\n", error->message);
-                g_error_free(error);
+                if (error != NULL) {
```
ejoerns, line 128: This should not be necessary as the method above must set 'error' in case it returns false.
This is one of the very basic requirements for correct GError handling in Glib. Or did you end up in cases where this was actually failing?

src/rauc-hawkbit-updater.c
```diff
@@ -155,8 +159,12 @@ int main(int argc, char **argv)
 
         struct config *config = load_config_file(config_file, &error);
         if (config == NULL) {
-                g_printerr("Loading config file failed: %s\n", error->message);
-                g_error_free(error);
+                if (error != NULL) {
```
ejoerns, line 162: The same as noted above applies here too. With the difference that `load_config_file()` can actually return NULL without setting GError properly. But this should be fixed in `load_config_file()`, not with a hack :smirk:

src/rauc-hawkbit-updater.c
Raphexion: I agree that this shouldn't be necessary. However, I can make is SEGFAULT by playing around with the config file. Currently there cases when error will end up being null. So this is just an extra layer of protection against.

src/rauc-hawkbit-updater.c
ejoerns: For sure, in general, it is not a bad idea to protect against NULL pointers.

But the check is wrong here as in case of a failure, the GError must contain a valid error and `NULL` is the indicator for 'no error'. Thus we use the condition actually indicating 'no error' to say 'there was an error we do not know about'. This is inconsistent and hides issues. [1]

This could be called a 'hack' or a 'bug', but anyway, it is no appropriate handling of GError. ;)

[1] https://developer.gnome.org/glib/stable/glib-Error-Reporting.html

Raphexion:
Fixed elsewhere.

PR #27, opened 2020-04-03
Fix Error Handling in config-file.c

### LINKED ISSUE #23: Segmentation fault caused by "config.conf" file
Hi,
when using the rauc-hawkbit-updater, we get a SIGSEGV. Segmentation Fault.
We noticed, that the updater crashes while trying to get the data from the “config.conf” file.
It seems, that the Boolean variables in the “config.conf” file are not parsed correctly and because of that the comments, which are declared with a “#” on the same line, are being ignored as comments and written into the variables.

Did anyone have the same Problem?

Thanks.

ejoerns:
This fixes the handling of GError's in config-file.c in many different ways.

As a result of this, no segmentation faults caused by unset errors should be triggered anymore and parsing of broken configuration should return valid error messages.

Fixes #23
Replaces #24

### DIFF
--- a/src/config-file.c
+++ b/src/config-file.c
@@ -211,8 +211,8 @@ struct config* load_config_file(const gchar* config_file, GError** error)
 
         if (config->timeout > 0 && config->connect_timeout > 0 && config->timeout < config->connect_timeout) {
                 g_set_error(error,
-                            1,                   // error domain
-                            5,                   // error code
+                            G_KEY_FILE_ERROR,                   // error domain
+                            G_KEY_FILE_ERROR_INVALID_VALUE,     // error code
                             "timeout should be greater than connect_timeout. Timeout: %ld, Connect timeout: %ld",
                             config->timeout,
                             config->connect_timeout
--- a/src/config-file.c
+++ b/src/config-file.c
@@ -46,12 +46,16 @@ void config_file_free(struct config *config)
 static gboolean get_key_string(GKeyFile *key_file, const gchar* group, const gchar* key, gchar** value, const gchar* default_value, GError **error)
 {
         gchar *val = NULL;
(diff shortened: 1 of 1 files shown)


src/config-file.c
```diff
@@ -76,38 +80,64 @@ static gboolean get_key_bool(GKeyFile *key_file, const gchar* group, const gchar
...
         guint key;
         gsize num_keys;
-        gchar **keys, *value;
+        gchar **keys;
 
         *hash = g_hash_table_new(g_str_hash, g_str_equal);
-        keys = g_key_file_get_keys(key_file, group, &num_keys, &error);
+        keys = g_key_file_get_keys(key_file, group, &num_keys, error);
+        if (keys == NULL)
+                return FALSE;
```
Raphexion, line 118: Do we need a g_set_error here (above return) since we return FALSE?

src/config-file.c
```diff
@@ -76,38 +80,64 @@ static gboolean get_key_bool(GKeyFile *key_file, const gchar* group, const gchar
...
-                value = g_key_file_get_value(key_file,
-                                             group,
-                                             keys[key],
-                                             &error);
+                gchar *value = g_key_file_get_value(key_file,
+                                                    group,
+                                                    keys[key],
+                                                    error);
+                if (value == NULL)
+                        return FALSE;
```
Raphexion, line 134: Do we need a g_set_error here (above return) since we return FALSE?

src/config-file.c
ejoerns: No, as `error` is set by `g_key_file_get_keys()` in this case.

src/config-file.c
ejoerns: same as above

PR #38, opened 2020-05-12
Add option to authenticate via client certificates

ctenruh-phytec:
Client certificate is send to the set Server.
Also added a new condition in "get_key_string" since before the function would return TRUE when "val" was given an empty string.
Edited the "rauc-hawkbit-updater.t" file accordingly so the tests won't fail.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,8 @@ Setup target (device) configuration file:
   target_name               = test-target
   auth_token                = bhVahL1Il1shie2aj2poojeChee6ahShu
   #gateway_token            = bhVahL1Il1shie2aj2poojeChee6ahShu
+  #client_cert              = /path/to/client_certificate.pem
+  #client_key               = /path/to/client_certificate.key
   bundle_download_location  = /tmp/bundle.raucb
   retry_wait                = 60
   connect_timeout           = 20
--- a/src/config-file.c
+++ b/src/config-file.c
@@ -240,6 +240,8 @@ Config* load_config_file(const gchar *config_file, GError **error)
         g_autoptr(GKeyFile) ini_file = NULL;
         gboolean key_auth_token_exists = FALSE;
         gboolean key_gateway_token_exists = FALSE;
+        gboolean key_client_cert_exists = FALSE;
+        gboolean key_client_key_exists = FALSE;
 
         g_return_val_if_fail(config_file, NULL);
         g_return_val_if_fail(error == NULL || *error == NULL, NULL);
@@ -259,11 +261,14 @@ Config* load_config_file(const gchar *config_file, GError **error)
         key_gateway_token_exists = get_key_string(ini_file, "client", "gateway_token",
                                                   &config->gateway_token, NULL, NULL);
         if (!key_auth_token_exists && !key_gateway_token_exists) {
-                g_set_error(error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_INVALID_VALUE,
-                            "Neither auth_token nor gateway_token is set in the config.");
-                return NULL;
-        }
-        if (key_auth_token_exists && key_gateway_token_exists) {
+                g_info("Neither auth_token nor gateway_token is set, using client certificates");
+                key_client_cert_exists = get_key_string(ini_file, "client", "client_cert", &config->client_cert, NULL, NULL);
+                key_client_key_exists = get_key_string(ini_file, "client", "client_key", &config->client_key, NULL, NULL);
+                if (!key_client_cert_exists || !key_client_key_exists) {
+                        g_set_error(error, 1, 4, "Neither a token nor client certificate are set!");
+                        return NULL;
+                }
+        } else if (key_auth_token_exists && key_gateway_token_exists) {
                 g_set_error(error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_INVALID_VALUE,
                             "Both auth_token and gateway_token are set in the config.");
                 return NULL;
@@ -321,6 +326,8 @@ void config_file_free(Config *config)
         g_free(config->tenant_id);
         g_free(config->auth_token);
         g_free(config->gateway_token);
+        g_free(config->client_cert);
+        g_free(config->client_key);
         g_free(config->bundle_download_location);
         if (config->device)
                 g_hash_table_destroy(config->device);
--- a/src/hawkbit-client.c
+++ b/src/hawkbit-client.c
@@ -262,6 +262,13 @@ static gboolean get_binary(const gchar *download_url, const gchar *file, gint64 
         curl_easy_setopt(curl, CURLOPT_URL, download_url);
         curl_easy_setopt(curl, CURLOPT_FOLLOWLOCATION, 1L);
         curl_easy_setopt(curl, CURLOPT_MAXREDIRS, 8L);
+
+        if (hawkbit_config->client_cert && hawkbit_config->client_key) {
+                curl_easy_setopt(curl, CURLOPT_SSLKEYTYPE, "PEM");
+                curl_easy_setopt(curl, CURLOPT_SSLCERT, hawkbit_config->client_cert);
+                curl_easy_setopt(curl, CURLOPT_SSLKEY, hawkbit_config->client_key);
+        }
+
         curl_easy_setopt(curl, CURLOPT_BUFFERSIZE, DEFAULT_CURL_DOWNLOAD_BUFFER_SIZE);
         curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, curl_write_to_file_cb);
         curl_easy_setopt(curl, CURLOPT_WRITEDATA, payload);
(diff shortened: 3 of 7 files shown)


Raphexion:
If you have time and if you find it appropriate, would you mind writing a a couple of lines about the client certificates in the README. We already have some text about the two different tokens. I think it could be helpful to new users of the updater.

ctenruh-phytec:
of course.

Raphexion:
@ctenruh-phytec Great work on the README :1st_place_medal:

ctenruh-phytec:
thanks :)

ctenruh-phytec:
Hello @bantu ,
thanks for your comment/review.
You are right, the private key is not send to the server, i will correct the README file.

README.md
```diff
@@ -121,3 +123,60 @@ In the RAUC hawkBit updater's configuration file it's called "gateway_token".
...
+	- disband the TLS connection
+	- check if sent client certificate is valid
+	- extract the common name and fingerprint of the client certificate
+	- forward the common name and fingerprint as HTTP headers to the
+	  hawkBit server
+When the hawkBit server receives the request it checks if:
+	- sent common name matches with the controller ID of the target
+	- sent fingerprint(s) matches the expected fingerprint(s) which is set
+	  in the system configuration settings of hawkBit
+The client certificate will only be sent if no tokens are set and a valid path
```
bantu, line 153: I think "sent" should be "used" here.

README.md
```diff
@@ -38,6 +38,8 @@ Setup target (device) configuration file:
   retry_wait                = 60
   connect_timeout           = 20
   timeout                   = 60
+  client_cert               =
```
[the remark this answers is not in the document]
ejoerns: I'd tend to add an example path, comment this out and place it below the other authentification options.
```suggestion
  # client_cert               = /path/to/client_cert.pem
```

README.md
```diff
@@ -101,8 +103,8 @@ Application Options:
 ```
 
 
-Authentication
---------------
+Authentication via tokens
```
ejoerns, line 106: Please use Capitalization in Headings
```suggestion
Authentication via Tokens
```

README.md
```diff
@@ -121,3 +123,60 @@ In the RAUC hawkBit updater's configuration file it's called "gateway_token".
...
+When the hawkBit server receives the request it checks if:
+	- sent common name matches with the controller ID of the target
+	- sent fingerprint(s) matches the expected fingerprint(s) which is set
+	  in the system configuration settings of hawkBit
+The client certificate will only be sent if no tokens are set and a valid path
+to a certificate and its key is given in the configuration file.
+
+Here an example of how the configuration file might look like:
+
+```
```
ejoerns, line 158: Use cfg as highlight type here.
```suggestion
``` cfg
```

src/config-file.c
```diff
@@ -41,14 +41,16 @@ void config_file_free(struct config *config)
...
+        g_free(config->client_key);
         g_hash_table_destroy(config->device);
 }
 
 static gboolean get_key_string(GKeyFile *key_file, const gchar* group, const gchar* key, gchar** value, const gchar* default_value, GError **error)
 {
         gchar *val = NULL;
         val = g_key_file_get_string(key_file, group, key, NULL);
-        if (val == NULL) {
+        if (val == NULL || !g_strcmp0(val, "")) {
```
ejoerns, line 53: g_strcmp0() returns an integer, thus I would perfer using it as `g_strcmp0(val, "") == 0` as otherwise, the first guess could be that we check for inequality.

Another more readable option would be to use `g_str_equals()` as we ensured val is != NULL already.

ejoerns (changes requested):
In general it looks like a good and relevant extension of the already-implemented authentication options for me.

I've also added some notes of what I've found while having a look on this.

We should also add some minimal tests for that (at least for option handling).

Please also note that the commit messages should be prefixed with the appropriate scope where applicable.

hongkongkiwi:
May I know when this is planned to be merged? It looks like all the needed changes are here?

README.md
```diff
@@ -121,3 +123,60 @@ In the RAUC hawkBit updater's configuration file it's called "gateway_token".
...
+
+Authentication via Certificates
+-------------------------------
+
+As can be seen in the system configuration settings of hawkBit, there is a
+third option to authenticate the targets. An "Allow targets to authenticate via
+a certificate authenticated by a reverse proxy" option. To use this
+authentication method a TLS reverse proxy server needs to be set up.
+The client and reverse proxy server perform a "SSL-handshake" that means the
+client validates the server certificate of the reverse proxy server with its
```
bantu, line 136: The server also validates the client's certificate? I.e. we are performing "mutual authentication"?

README.md
```diff
@@ -121,3 +123,60 @@ In the RAUC hawkBit updater's configuration file it's called "gateway_token".
...
+list of trusted certificates.
```
bantu, line 137: How does the client know which certificate authorities (CAs) to trust? Does it make sense to add an option for explicitly defining a directory with trusted CA certificates (or a single file containing multiple certificates)?

bantu: On second thought: This also applies to the "Authentication with Tokens" scenario.

bantu:
@hongkongkiwi It looks like this PR needs to be rebased (or merged) to resolve some conflicts.

I'd like to see this merged too. Anything in particular I can help with?

README.md
ctenruh-phytec: Yes, the client request has to contain the client certificate, which the server has to verify.

README.md
ctenruh-phytec: I think most devices already have a file which contains the trusted certificates.

ctenruh-phytec:
i noticed you removed the Authentication Methods from the README so i also removed my entry for the verification via certificates.
I applied all the changes requested, does it need anything else to be merged?

Bastian-Krause:
@ctenruh-phytec The authentication section from the README was moved to `docs/using.rst` with #74 (422d90d4). Maybe add it there?

ctenruh-phytec:
moved the description to `docs/using.rst` . Thanks @Bastian-Krause

src/config-file.c
```diff
@@ -272,11 +274,14 @@ Config* load_config_file(const gchar *config_file, GError **error)
...
-        }
-        if (key_auth_token_exists && key_gateway_token_exists) {
+                g_info("Neither auth_token nor gateway_token is set, using client certificates");
+                key_client_cert_exists = get_key_string(ini_file, "client", "client_cert", &config->client_cert, NULL, NULL);
+                key_client_key_exists = get_key_string(ini_file, "client", "client_key", &config->client_key, NULL, NULL);
+                if (!key_client_cert_exists || !key_client_key_exists) {
+                        g_set_error(error, 1, 4, "Neither a token nor client certificate are set!");
+                        return NULL;
+                }
+        } else if (key_auth_token_exists && key_gateway_token_exists) {
```
Bastian-Krause, line 284: Could you please read these config options outside of these `if` and then add checks for all combinations that are not allowed? This should make the logic simpler here.

src/hawkbit-client.c
```diff
@@ -276,6 +276,13 @@ static gboolean get_binary(const gchar *download_url, const gchar *file, gint64
         curl_easy_setopt(curl, CURLOPT_URL, download_url);
         curl_easy_setopt(curl, CURLOPT_FOLLOWLOCATION, 1L);
         curl_easy_setopt(curl, CURLOPT_MAXREDIRS, 8L);
+
+        if (hawkbit_config->client_cert && hawkbit_config->client_key) {
+                curl_easy_setopt(curl, CURLOPT_SSLKEYTYPE, "PEM");
+                curl_easy_setopt(curl, CURLOPT_SSLCERT, hawkbit_config->client_cert);
+                curl_easy_setopt(curl, CURLOPT_SSLKEY, hawkbit_config->client_key);
+        }
+
```
Bastian-Krause, line 285: This belongs in `set_auth_curl_header()` which is used for `get_binary()` as well as `rest_request()`.

src/hawkbit-client.c
```diff
@@ -382,6 +389,13 @@ static gboolean rest_request(enum HTTPMethod method, const gchar *url,
         // set up CURL options
         set_default_curl_opts(curl);
         curl_easy_setopt(curl, CURLOPT_URL, url);
+
+        if (hawkbit_config->client_cert && hawkbit_config->client_key) {
+                curl_easy_setopt(curl, CURLOPT_SSLKEYTYPE, "PEM");
+                curl_easy_setopt(curl, CURLOPT_SSLCERT, hawkbit_config->client_cert);
+                curl_easy_setopt(curl, CURLOPT_SSLKEY, hawkbit_config->client_key);
+        }
+
```
Bastian-Krause, line 398: See above.

test/rauc-hawkbit-updater.t
```diff
@@ -61,7 +61,7 @@ test_expect_success "rauc-hawkbit-updater register and check (invalid auth token
 
 test_expect_success "rauc-hawkbit-updater register and check (no security tokens)" "
   cp $SHARNESS_TEST_DIRECTORY/test-config-no-security-tokens.conf . &&
-  echo 'Loading config file failed: Neither auth_token nor gateway_token is set in the config.' > expected_out &&
+  echo 'Loading config file failed: Neither a token nor client certificate are set!' > expected_out &&
```
Bastian-Krause, line 64: The bash tests are gone since #83. It would be great to have an actual test case for this with a reverse proxy. We already use nginx for some test cases (see test/conftest.py).

Bastian-Krause (changes requested):
We also need documentation for the new options in doc/reference.rst.

ejoerns:
@ctenruh-phytec No worries, same here ;)

Bastian-Krause:
Superseded by #166.

PR #39, opened 2020-05-12
Add a reboot option

ctenruh-phytec:
To reboot system after an update set "post_update_reboot" variable in config
file accordingly.

### DIFF
--- a/src/config-file.c
+++ b/src/config-file.c
@@ -34,6 +34,7 @@ static const gint DEFAULT_TIMEOUT         = 60;     // 1 min.
 static const gint DEFAULT_RETRY_WAIT      = 5 * 60; // 5 min.
 static const gboolean DEFAULT_SSL         = TRUE;
 static const gboolean DEFAULT_SSL_VERIFY  = TRUE;
+static const gboolean DEFAULT_REBOOT      = FALSE;
 static const gchar* DEFAULT_LOG_LEVEL     = "message";
 
 /**
@@ -307,6 +308,9 @@ Config* load_config_file(const gchar *config_file, GError **error)
         if (!get_key_string(ini_file, "client", "log_level", &val, DEFAULT_LOG_LEVEL, error))
                 return NULL;
         config->log_level = log_level_from_string(val);
+
+        if (!get_key_bool(ini_file, "client", "post_update_reboot", &config->post_update_reboot, DEFAULT_REBOOT, error))
+                return NULL;
 
         if (config->timeout > 0 && config->connect_timeout > 0 &&
             config->timeout < config->connect_timeout) {
--- a/src/hawkbit-client.c
+++ b/src/hawkbit-client.c
@@ -46,6 +46,7 @@
 #include <libgen.h>
 #include <bits/types/struct_tm.h>
 #include <gio/gio.h>
+#include <sys/reboot.h>
 
 #include "json-helper.h"
 #ifdef WITH_SYSTEMD
@@ -544,6 +545,13 @@ gboolean install_complete_cb(gpointer ptr)
                 if (result->install_success) {
                         g_message("Software bundle installed successful.");
                         feedback(feedback_url, action_id, "Software bundle installed successful.", "success", "closed", NULL);
+                        if (hawkbit_config->post_update_reboot) {
+                                process_deployment_cleanup();
+                                sync();
+                                if (reboot(RB_AUTOBOOT) < 0) {
+                                        g_critical("Failed to reboot: %s", g_strerror(errno));
+                                }
+                        }
                 } else {
                         g_critical("Failed to install software bundle.");
                         feedback(feedback_url, action_id, "Failed to install software bundle.", "failure", "closed", NULL);
--- a/test/test-config.conf
+++ b/test/test-config.conf
@@ -10,6 +10,7 @@ retry_wait                = 60
 connect_timeout           = 20
 timeout                   = 60
 log_level                 = debug
+post_update_reboot        = false
 
 [device]
 product                   = Terminator
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ Setup target (device) configuration file:
   connect_timeout           = 20
   timeout                   = 60
   log_level                 = debug
+  post_update_reboot        = false
 
   [device]
   product                   = Terminator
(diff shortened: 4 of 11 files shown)


hmenn:
I've applied changes on our target but I've faced with a problem. Current rauc-hawkbit-updater service user is rauc-hawkbit and if this user doesn't have admin privileges, reboot call doesn't work. I've removed user, group config from service file to use this functionality. Do you have another suggestions? We might give reboot ability to rauc-hawkbit user.

ex. [stackoverflow](https://unix.stackexchange.com/a/85665)

prevas-lkmi:
If you are using systemd, I think you can give the rauc-hawkbit user reboot capability by using CapabilityBoundingSet with CAP_SYS_BOOT

Something like this

>[Service]
>User=rauc-hawkbit
>Group=rauc-hawkbit
>CapabilityBoundingSet=CAP_SYS_BOOT
>...

I have not tested it.

hmenn:
I've integrated this functionality with our current systems and works awesome. CAP_SYS_BOOT flag is required. This should be good feature.

ctenruh-phytec:
Granted the user "rauc-hawkbit" reboot capabilities with `AmbientCapabilities=CAP_SYS_BOOT`

src/config-file.c
```diff
@@ -44,6 +45,7 @@ void config_file_free(struct config *config)
         g_free(config->auth_token);
         g_free(config->gateway_token);
         g_free(config->bundle_download_location);
+        g_free(config->post_update_reboot);
```
prevas-lkmi, line 48: post_update_reboot is not a pointer. Nothing to free. Remove this line

prevas-lkmi (changes requested):

src/config-file.c
ctenruh-phytec: yes of course, thanks.

src/hawkbit-client.c
```diff
@@ -537,6 +538,10 @@ gboolean install_complete_cb(gpointer ptr)
                 if (result->install_success) {
                         g_message("Software bundle installed successful.");
                         feedback(feedback_url, action_id, "Software bundle installed successful.", "success", "closed", NULL);
+                        if (hawkbit_config->post_update_reboot) {
+                                sync();
```
mschwan-phytec, line 542: Maybe it is worth executing `process_deployment_cleanup()` before syncing and rebooting to remove any leftover files that have been downloaded.

ctenruh-phytec: yes, in case someone changes the `bundle_download_location` this would be useful. Thanks.

hmenn:
@prevas-lkmi It seems changed request was applied in previous commits. Could you please check it again? Should we merge reboot functionality in mainline?

ctenruh-phytec:
Hello, i just wanted to follow up on this issue. Did you have the chance to review the changes?

test/test-config.conf
```diff
@@ -10,6 +10,7 @@ retry_wait                = 60
 connect_timeout           = 20
 timeout                   = 60
 log_level                 = debug
+post_update_reboot        = false
```
SijmenHuizenga, line 13: Add this option to the config file in the readme as well?

ejoerns:
@ctenruh-phytec We discussed this last week internally to figure out what our opinion on this actually is.

The thing is that I do not really like the way we reboot here as this will kill all applications etc. immediately and will probably cause data loss etc.

The preferred way of rebooting would be to either

* send a reboot request to the application (that then performs the actual reboot depending on what its current state is),
* or at least send out a notification to the applications and all services that the system will reboot soon and give them some time to terminate gracefully and save data, persist logs, etc.

While the first one would require some additional interfaces being defined, the latter one is what most systems provide. A simple `reboot` or `systemctl reboot` would give the applications time to shut down. Unfortunately, there is no easy way we would know of that allows performing this as a non-root user. Ideas with using Policykit etc. exist but are non-trivial and require more to set up in the system.

Thus, yes, if we want to enable automatic rebooting out of rauc-hawkbit-updater, and in general I agree we want to, then currently, doing it the 'hard-way' seems to be the only option for non-root services.

Thus things left to be done here are rebasing onto the current master and adding a check on the return value of the `reboot()` call to allow logging when rebooting failed for any reason.
When we have some documentation, it's pretty sure good to also note what we can expect from that option and what not (graceful shutdown).

src/hawkbit-client.c
```diff
@@ -544,6 +545,13 @@ gboolean install_complete_cb(gpointer ptr)
                 if (result->install_success) {
                         g_message("Software bundle installed successful.");
                         feedback(feedback_url, action_id, "Software bundle installed successful.", "success", "closed", NULL);
+                        if (hawkbit_config->post_update_reboot) {
+                                process_deployment_cleanup();
+                                sync();
+                                if(reboot(RB_AUTOBOOT) < 0){
+					g_warning("Reboot failed - reboot() call returned: -1");
+				}
```
mschwan-phytec, line 553: 1. Use space instead of tabs to be consistent with the style of the existing code.
2. Add spacing around the `if`'s parentheses.
3. Instead of printing that `reboot()` may have returned `-1` in any case of failure, you should parse the value of `errno` here to give a more meaningful result.
4. IMHO if `reboot()` fails, it should be considered a critical failure and not just a warning, because the user clearly specified to reboot after an update but the program could not execute as expected. I am open to different opinions on this one though.

mschwan-phytec (changes requested):

src/hawkbit-client.c
ejoerns: @ctenruh-phytec For the style issues, you can easily have a look at the 'uncrustify' test result: https://github.com/rauc/rauc-hawkbit-updater/pull/39/checks?check_run_id=1777796976 You can also run 'uncrustify' manually from the repo by calling `./uncrustify.sh`.

ejoerns:
The test build should be fixed now, thus please also rebase onto master.

src/hawkbit-client.c
ctenruh-phytec: thanks for the review.

ejoerns (approved):

prevas-lkmi (approved):
LGTM

PR #46, opened 2020-08-16
add poll interval option

hmenn:
* Previous polling interval was 1second and it was not configurable
  via config file. For some specific reasons like network usage, user
  might want to increase polling interval


ejoerns:
@hmenn thanks for your contribution. But the timeout interval you made configurable is only the interval the callback function (`hawkbit_pull_cb`) is called with. If you refer to network usage, then the `sleep_time` variable is the relevant polling interval.
This comes with a default but is sett according to the polling interval configured and announced by hawkbit (`config.polling.sleep`).

hmenn:
@ejoerns Thanks a lot. I saw that configuration from my server side. I'm closing pr.

PR #47, opened 2020-08-17
hawkbit-client: Fix memory leak on debug messages

### LINKED ISSUE #45: hawkbit-client.c: memory leak on debug messages
hawkbit-client. c is making use of `json_to_string()` for debugging messages ([1](https://github.com/rauc/rauc-hawkbit-updater/blob/master/src/hawkbit-client.c#L641), [2](https://github.com/rauc/rauc-hawkbit-updater/blob/master/src/hawkbit-client.c#L750)). However, this is provoking a small leak as said function returns a `char *` that must be released manually. 

A possible fix could be:

```c
char *str = json_to_string(resp_root, TRUE);
g_debug("Deployment response: %s\n", str);
free(str);
```

If this approach is OK, we can proceed to generate a PR :)

rfuentess:
This fixes #45 (memory leak on debug messages). 

NOTE: Although I detected the leak, it was @hpatriarche  who truly was able to trace it to its source. Therefore, I'm just the committer.

### DIFF
--- a/src/hawkbit-client.c
+++ b/src/hawkbit-client.c
@@ -598,6 +598,7 @@ static gboolean process_deployment(JsonNode *req_root, GError **error)
 {
         GError *ierror = NULL;
         struct artifact *artifact = NULL;
+        g_autofree gchar *str = NULL;
 
         if (action_id) {
                 g_warning("Deployment is already in progress...");
@@ -638,7 +639,9 @@ static gboolean process_deployment(JsonNode *req_root, GError **error)
                 goto proc_error;
         }
         JsonNode *resp_root = json_parser_get_root(json_response_parser);
-        g_debug("Deployment response: %s\n", json_to_string(resp_root, TRUE));
+
+        *str = json_to_string(resp_root, TRUE);
+        g_debug("Deployment response: %s\n", str);
 
         JsonArray *json_chunks = json_get_array(resp_root, "$.deployment.chunks");
         if (json_chunks == NULL || json_array_get_length(json_chunks) == 0) {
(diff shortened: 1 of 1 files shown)


ejoerns:
@rfuentess It does not really count who discovered the fix. If you did the patch, I'd consider you as the author. If @hpatriarche created the patch and you only submit it, then it's fine, but the Signed-off-by line does not reflect this.

There is nothing that speaks against referring him in the commit message (like with a "Suggested-by: .." line as the Linux kernel does it). Also note, that despite I suggested the actual content the patch now has, I wouldn't see me as the author ;)

rfuentess:
@ejoerns  Fair enough. I saw the sign-off too late. I'll amend the commit message to do it as suggested. But, before doing that. It seems that Travis is not happy at all:

```
/home/travis/build/rauc/rauc-hawkbit-updater/src/hawkbit-client.c:641:26: note: 
      jump bypasses initialization of variable with __attribute__((cleanup))
        g_autofree char *str = json_to_string(resp_root, TRUE);
```

Do I need to do something special ? 
P.s. Travis also complains about a goto previous to this line.  I'm not sure if they are related.

ejoerns:
@rfuentess Oh yes, it cannot deterministically initialize the str variable we use for the cleanup foo. The solution would be put the initialization of the variable before the `goto` statement so that we end up with something like

``` c
        g_autofree char *str = NULL;

        [...]
            goto proc_error;
        [...]
        str = json_to_string(resp_root, TRUE);
        g_debug("Deployment response: %s\n", str);
```

However, this is a flaw that affects the entire module code...

rfuentess:
Done. I amended the commit message and now Travis is happy

src/hawkbit-client.c
```diff
@@ -598,6 +598,7 @@ static gboolean process_deployment(JsonNode *req_root, GError **error)
 {
         GError *ierror = NULL;
         struct artifact *artifact = NULL;
+        g_autofree gchar *str;
```
ejoerns, line 601: Note that in order to have autocleanup working correctly, this needs to be initialized to NULL.
```suggestion
        g_autofree gchar *str = NULL;
```

rfuentess: :man_facepalming:   Sorry.  I saw  the implementation in other points and then... I forget to put the NULL :man_facepalming: :man_facepalming: .  Do I push a fixup with it or I directly do a forced push ?

ejoerns: just do a force-push

PR #48, opened 2020-08-21
Fix wrong assignment of char* to char

ejoerns:
This was introduced and overseen in #47.

However, this clearly triggered a compiler warning but that was not discovered. For that reason, this PR also includes building the code with `-Werror` in Travis so that all compiler warnings a treated as errors and will let the build fail visibly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: c
 
-dist: xenial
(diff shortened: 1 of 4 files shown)


prevas-lkmi:
@ejoerns Good catch. LGTM

PR #50, opened 2020-08-26
Fix free storage space calculation overflowing

fshrOSB:
The number of available bytes of storage calculated can overflow when there is too much free space.

On a machine where long is only 32 bits in size the overflow occurs with more than about 2 GB available.
The `get_available_space()` function then reports a negative value and the rollout fails with "Not enough free space" e.g..
`Error: Not enough free space. File size: 71264324. Free space: -94932992`

### DIFF
--- a/src/hawkbit-client.c
+++ b/src/hawkbit-client.c
@@ -75,7 +75,7 @@ static long last_run_sec = 0;
  * @param[in] path Path
  * @return If error -1 else free space in bytes
  */
-static long get_available_space(const char* path, GError **error)
+static goffset get_available_space(const char* path, GError **error)
 {
         struct statvfs stat;
         g_autofree gchar *npath = g_strdup(path);
@@ -85,8 +85,9 @@ static long get_available_space(const char* path, GError **error)
                 g_set_error(error, G_FILE_ERROR, G_FILE_ERROR_FAILED, "Failed to calculate free space: %s", g_strerror(errno));
                 return -1;
         }
+
         // the available free space is f_bsize * f_bavail
-        return stat.f_bsize * stat.f_bavail;
+        return (goffset) stat.f_bsize * (goffset) stat.f_bavail;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


ejoerns (changes requested):
@fshrOSB Thanks for your contribution! Please note that you can simply force-push to your PR branch to update the commit instead of closing and opening another one.

Otherwise, it looks like a reasonable fix and good catch. :+1: 

Please also add a proper prefix to the commit message (e.g. `src/hawkbit-client: ...`) and fix the minor typo ('storage').

ejoerns (commented):
It basically looks ok for me (no idea why Travis is not responding here despite the build looks fine), theonly thing I came across when having a look at the corresponding code in RAUC is that we might want to use the `goffset` type here, too. Which is explicitly meant to be used for file sizes and offsets. Effectively it is the same as `gint64`.
Any opinions?

I also wondered whether glib has something built-in for getting free disk space. Having a short look into the manual shows there `is g_file_query_info()` and the `G_FILE_ATTRIBUTES`, but maybe this is a bit heavier than required.

src/hawkbit-client.c
```diff
@@ -867,3 +868,4 @@ int hawkbit_start_service_sync()
 
         return res;
 }
+
```
mschwan-phytec, line 871: Please remove the added extra line at the end of the file. Everything else looks good to me.

src/hawkbit-client.c
```diff
@@ -868,4 +868,3 @@ int hawkbit_start_service_sync()
 
         return res;
 }
-
```
mschwan-phytec: Thanks for fixing this. You should squash this commit into your previous one, as the newline was introduced by your previous commit of this PR.

mschwan-phytec:
Just a quick tip regarding you commit messages: You may want to keep the line width to a max. 72 characters. I don't think this is a strict requirement for this project but is good practice to increase readability.

ejoerns:
I'd really do this in a single commit (instead of moving form `long` -> `int64` -> `goffset`), but otherwise, it looks fine for me.

fshrOSB:
> It basically looks ok for me (no idea why Travis is not responding here despite the build looks fine), theonly thing I came across when having a look at the corresponding code in RAUC is that we might want to use the `goffset` type here, too. Which is explicitly meant to be used for file sizes and offsets. Effectively it is the same as `gint64`.
> Any opinions?
> 
> I also wondered whether glib has something built-in for getting free disk space. Having a short look into the manual shows there `is g_file_query_info()` and the `G_FILE_ATTRIBUTES`, but maybe this is a bit heavier than required.

I changed the type for the available space from `gint64` to `goffset`.

The `size` element of the `artifact` struct, to which the available space is compared to, should probably also switch to the `goffset` type in the future.
https://github.com/rauc/rauc-hawkbit-updater/blob/cb55819522be9566e074efd2d09d425a2f3ebc52/include/hawkbit-client.h#L73-L84

ejoerns (approved):
Thanks for your changes, looks fine for me now!

PR #52, opened 2020-09-01
src/hawkbit-client: fix potential gchar leakage in process_deployment()

ejoerns:
In case `regex_groups()` call fails, `deployment` will not be freed.
Simply use autopointer here to fix this.


prevas-lkmi (approved):
LGTM

PR #53, opened 2020-09-02
src/hawkbit-client: var args support for build_api_url()

ejoerns:
This simplifies usage of `build_api_url()` that is always used for
constructing API strings with g_strdup_printf()`.

As the memory allocated was not always freed but sometimes directly
passed this also fixes some memory leakages.

The `__attribute__` tells the compiler to validate the 'path' argument to
be in the right format for the provided arguments.


prevas-lkmi (approved):
LGTM

PR #54, opened 2020-09-03
src/hawkbit-client: use long for CURLINFO_RESPONSE_CODE

ejoerns:
libcurl documentation for `CURLINFO_RESPONSE_CODE` notes:

> Pass a pointer to a long to receive the last received HTTP, FTP or
> SMTP response code.

Passing only an integer instead can lead to a segfault as unrelated
memory is overridden.


prevas-lkmi (approved):
LGTM

PR #56, opened 2020-09-04
src/hawkbit-client: allow following HTTP redirects

### LINKED ISSUE #40: Curl Follow Location
Hi,

In our version, we added 
`curl_easy_setopt(curl, CURLOPT_FOLLOWLOCATION, 1L);`

is it an option that make sense by default ?

ejoerns:
There is no special reason for not doing this and most tools using curl
have this enabled (including rauc) thus we allow redirects here too,
but with a limit of 5 redirects which seems to be a common default.

Fixes #40


prevas-lkmi (approved):
LGTM

PR #63, opened 2020-12-04
json-helper: documentation, refactoring and bug fixes

Bastian-Krause:
This is the first of a series of PRs to refactor and clean up the code base, get rid of bugs (memory leaks, use-after-frees, race conditions), simplify functions and document them.

In this iteration I focused on the `json-helper` module (which required some minor changes in `hawkbit-client`).

### DIFF
--- a/include/json-helper.h
+++ b/include/json-helper.h
@@ -45,7 +45,17 @@ gchar* json_get_string(JsonNode *json_node, const gchar *path, GError **error);
  * @return gint64, integer value, 0 on error (error set)
  */
 gint64 json_get_int(JsonNode *json_node, const gchar *path, GError **error);
-JsonArray* json_get_array(JsonNode *json_node, const gchar *path);
+
+/**
+ * @brief Get the JsonArray inside the first JsonNode element matching path in json_node.
+ *
+ * @param[in]  json_node JsonNode to evaluate expression on
+ * @param[in]  path      JSONPath expression
+ * @param[out] error     Error
+ * @return JsonArray*, array (must be freed), NULL on error (error set)
+ */
+JsonArray* json_get_array(JsonNode *json_node, const gchar *path, GError **error);
+
 gboolean json_contains(JsonNode *root, gchar *key);
 
 #endif // __JSON_HELPER_H__
--- a/include/json-helper.h
+++ b/include/json-helper.h
@@ -26,7 +26,15 @@
 #include <glib/gtypes.h>
 #include <json-glib/json-glib.h>
 
-gchar* json_get_string(JsonNode *json_node, const gchar *path);
+/**
+ * @brief Get the string inside the first JsonNode element matching path in json_node.
+ *
+ * @param[in]  json_node JsonNode to evaluate expression on
+ * @param[in]  path      JSONPath expression
+ * @param[out] error     Error
+ * @return gchar*, string value (must be freed), NULL on error (error set)
+ */
+gchar* json_get_string(JsonNode *json_node, const gchar *path, GError **error);
 gint64 json_get_int(JsonNode *json_node, const gchar *path);
 JsonArray* json_get_array(JsonNode *json_node, const gchar *path);
 gboolean json_contains(JsonNode *root, gchar *key);
--- a/include/json-helper.h
+++ b/include/json-helper.h
@@ -56,6 +56,13 @@ gint64 json_get_int(JsonNode *json_node, const gchar *path, GError **error);
  */
 JsonArray* json_get_array(JsonNode *json_node, const gchar *path, GError **error);
 
+/**
+ * @brief Check if the given path matches an element in json_node.
+ *
+ * @param[in] json_node JsonNode to evaluate expression on
+ * @param[in] path      JSONPath expression
+ * @return gboolean, TRUE if path matches an element, FALSE otherwise
+ */
 gboolean json_contains(JsonNode *root, gchar *key);
 
 #endif // __JSON_HELPER_H__
--- a/include/json-helper.h
+++ b/include/json-helper.h
@@ -35,7 +35,16 @@
  * @return gchar*, string value (must be freed), NULL on error (error set)
  */
 gchar* json_get_string(JsonNode *json_node, const gchar *path, GError **error);
-gint64 json_get_int(JsonNode *json_node, const gchar *path);
+
+/**
+ * @brief Get the integer inside the first JsonNode element matching path in json_node.
+ *
+ * @param[in]  json_node JsonNode to evaluate expression on
+ * @param[in]  path      JSONPath expression
+ * @param[out] error     Error
+ * @return gint64, integer value, 0 on error (error set)
+ */
+gint64 json_get_int(JsonNode *json_node, const gchar *path, GError **error);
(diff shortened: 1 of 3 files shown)


Bastian-Krause:
Force-pushed:
- move doc strings of non-static functions to header file

src/json-helper.c
```diff
@@ -27,6 +27,51 @@
...
+ * @return JsonNode*, matching JsonNode element (must be freed), NULL on error
+ */
+static JsonNode* json_get_first_matching_element(JsonNode *json_node,
+                                                 const gchar *path,
+                                                 GError **error)
+{
+        g_autoptr(JsonNode) match = NULL, node = NULL;
+        JsonArray *arr = NULL;
+
+        g_return_val_if_fail(error == NULL || *error == NULL, NULL);
```
ejoerns, line 46: please do function argument checks in the order they appear in the function signature

src/json-helper.c
```diff
@@ -72,18 +72,28 @@ static JsonNode* json_get_first_matching_element(JsonNode *json_node,
...
+gchar* json_get_string(JsonNode *json_node, const gchar *path, GError **error)
 {
-        char *res_str = NULL;
-        JsonNode *result = json_path_query(path, json_node, NULL);
-        if (result) {
-                JsonArray *results = json_node_get_array(result);
-                if (json_array_get_length( results ) > 0) {
-                        res_str = g_strdup(json_array_get_string_element(results, 0));
-                }
+        g_autofree char *res_str = NULL;
```
ejoerns, line 77: If we use glib cleanup here, we should also change to glib type here (`gchar`).

src/json-helper.c
```diff
@@ -72,18 +72,28 @@ static JsonNode* json_get_first_matching_element(JsonNode *json_node,
...
+        g_autoptr(JsonNode) result = NULL;
+
+        g_return_val_if_fail(error == NULL || *error == NULL, NULL);
```
ejoerns, line 80: same here with arg order

src/json-helper.c
```diff
@@ -96,18 +96,26 @@ gchar* json_get_string(JsonNode *json_node, const gchar *path, GError **error)
...
-        gint64 res_int = -1;
-        JsonNode *result = json_path_query(path, json_node, NULL);
-        if (result) {
-                JsonArray *results = json_node_get_array(result);
-                if (json_array_get_length( results ) > 0) {
-                        res_int = json_array_get_int_element(results, 0);
-                }
+        g_autoptr(JsonNode) result = NULL;
+
+        g_return_val_if_fail(error == NULL || *error == NULL, -1);
```
ejoerns, line 103: glib functions return a `0` to indicate an error. I'd like to stay consistent with this.

I do not really like the violation of common glib error handling that we do not have a return code clearly indicating an error, but glib functions do it the same way and it would otherwise require an output argument, thus ok.

Maybe we should add a note in the signature documentation that only a `0` with GError set is an error indicator?

include/json-helper.h
```diff
@@ -59,6 +59,14 @@ gint64 json_get_int(JsonNode *json_node, const gchar *path, GError **error);
  */
 JsonArray* json_get_array(JsonNode *json_node, const gchar *path,
                 GError **error);
+
+/**
+ * @brief Check if the given path matches an element in json_node.
+ *
+ * @param[in]  json_node JsonNode to query
+ * @param[in]  path      Query path
```
ejoerns, line 67: This input here in detail would be a "JSONPath expression". glib methods call this variable "expression"

src/json-helper.c
```diff
@@ -149,10 +149,22 @@ JsonArray* json_get_array(JsonNode *json_node, const gchar *path,
...
-        return result;
+        g_autoptr(GError) error = NULL;
+        g_autoptr(JsonNode) node = NULL;
+
+        g_return_val_if_fail(json_node, FALSE);
+        g_return_val_if_fail(path, FALSE);
+
+        node = json_path_query(path, json_node, &error);
+        if (!node) {
+                g_warning("%s", error->message);
```
ejoerns, line 162: Maybe add a line

     /* failed to compile expression to JSONPath */

to indicate that the error here is only triggered when the 'path' argument is wrong, but not when the query did not fin anyhting (I myself had to look up that json_path_query() acutally always returns an array).

src/json-helper.c
```diff
@@ -27,6 +27,51 @@
...
+
+/**
+ * @brief Get the first JsonNode element matching path in json_node.
+ *
+ * @param[in]  json_node JsonNode to query
+ * @param[in]  path      Query path
+ * @param[out] error     Error
+ * @return JsonNode*, matching JsonNode element (must be freed), NULL on error
+ */
+static JsonNode* json_get_first_matching_element(JsonNode *json_node,
```
ejoerns, line 39: Note that introducing a static non-used function will trigger a warning message for this commit. Maybe ok if not testing with -Werror on.

ejoerns (changes requested):
Good work!

Some minor comments inline from my side before I am fully satisfied with this.

Some minor addition: Minor type in the first commit: "inside the **first** matching json"

src/json-helper.c
Bastian-Krause: Done.

src/json-helper.c
Bastian-Krause: Done.

src/json-helper.c
Bastian-Krause: Done.

src/json-helper.c
Bastian-Krause: Force-pushed 0 return on error and mentioned it in the doc string.

include/json-helper.h
Bastian-Krause: Right, fixed.

src/json-helper.c
Bastian-Krause: Done.

src/json-helper.c
Bastian-Krause: True, but that shouldn't be a problem. I'd like to keep the commits separated as they are now.

ejoerns: Ok

src/hawkbit-client.c
```diff
@@ -682,7 +682,13 @@ static gboolean process_deployment(JsonNode *req_root, GError **error)
         artifact = g_new0(struct artifact, 1);
         artifact->version = json_get_string(json_chunk, "$.version", NULL);
         artifact->name = json_get_string(json_chunk, "$.name", NULL);
-        artifact->size = json_get_int(json_artifact, "$.size");
+        artifact->size = json_get_int(json_artifact, "$.size", NULL);
```
ejoerns, line 685: Is there any advantage in adding this check here already? It is not noted in the commit and it checks for the return value of `json_get_int`, which not necessarily means a parsing error but could also only return the valid value `0`.

Bastian-Krause: True, I'll drop this check here. The (to be created) `hawkbit-client` refactoring PR will then contain a check for `0 && error`.

Bastian-Krause: Done.

ejoerns (approved):

PR #65, opened 2020-12-04
config-file: documentation, refactoring and bug fixes

Bastian-Krause:
This is the third of a series of PRs to refactor and clean up the code base, get rid of bugs (memory leaks, use-after-frees, race conditions), simplify functions and document them.

In this iteration I focused on the `config-file` module (which required some minor changes in `hawkbit-client`).

### DIFF
--- a/src/config-file.c
+++ b/src/config-file.c
@@ -36,17 +36,6 @@ static const gboolean DEFAULT_SSL         = TRUE;
 static const gboolean DEFAULT_SSL_VERIFY  = TRUE;
 static const gchar * DEFAULT_LOG_LEVEL    = "message";
 
-void config_file_free(struct config *config)
-{
-        g_free(config->hawkbit_server);
-        g_free(config->controller_id);
-        g_free(config->tenant_id);
-        g_free(config->auth_token);
-        g_free(config->gateway_token);
-        g_free(config->bundle_download_location);
-        g_hash_table_destroy(config->device);
-}
-
 static gboolean get_key_string(GKeyFile *key_file, const gchar* group, const gchar* key, gchar** value, const gchar* default_value, GError **error)
 {
         gchar *val = NULL;
@@ -168,9 +157,9 @@ static GLogLevelFlags log_level_from_string(const gchar *log_level)
         }
 }
 
-struct config* load_config_file(const gchar* config_file, GError** error)
+Config* load_config_file(const gchar* config_file, GError** error)
 {
-        struct config *config = g_new0(struct config, 1);
+        Config *config = g_new0(Config, 1);
 
         gint val_int;
         g_autofree gchar *val = NULL;
@@ -235,4 +224,20 @@ struct config* load_config_file(const gchar* config_file, GError** error)
         }
 
         return config;
+}
+
+void config_file_free(Config *config)
+{
+        if (!config)
+                return;
+
+        g_free(config->hawkbit_server);
+        g_free(config->controller_id);
+        g_free(config->tenant_id);
+        g_free(config->auth_token);
+        g_free(config->gateway_token);
+        g_free(config->bundle_download_location);
+        if (config->device)
+                g_hash_table_destroy(config->device);
+        g_free(config);
 }
--- a/src/config-file.c
+++ b/src/config-file.c
@@ -36,22 +36,42 @@ static const gboolean DEFAULT_SSL         = TRUE;
 static const gboolean DEFAULT_SSL_VERIFY  = TRUE;
 static const gchar * DEFAULT_LOG_LEVEL    = "message";
 
-static gboolean get_key_string(GKeyFile *key_file, const gchar* group, const gchar* key, gchar** value, const gchar* default_value, GError **error)
+/**
+ * @brief Get string value from key_file for key in group, optional default_value can be specified
+ * that will be used in case key is not found in group.
+ *
+ * @param[in]  key_file      GKeyFile to look value up
+ * @param[in]  group         A group name
+ * @param[in]  key           A key
+ * @param[out] value         Output string value
+ * @param[in]  default_value String value to return in case no value found, or NULL (not found
+ *                           leads to error)
+ * @param[out] error         Error
+ * @return TRUE if found, TRUE if not found and default_value given, FALSE otherwise (error is set)
+ */
+static gboolean get_key_string(GKeyFile *key_file, const gchar *group, const gchar *key,
+                               gchar **value, const gchar *default_value, GError **error)
 {
-        gchar *val = NULL;
-        val = g_key_file_get_string(key_file, group, key, NULL);
-        if (val == NULL) {
-                if (default_value != NULL) {
+        g_autofree gchar *val = NULL;
+
+        g_return_val_if_fail(key_file, FALSE);
+        g_return_val_if_fail(group, FALSE);
+        g_return_val_if_fail(key, FALSE);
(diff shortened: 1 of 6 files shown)


Bastian-Krause:
Force-pushed:
- move doc string of non-static function to header file
- doc string formatting fixes

Bastian-Krause:
Force-pushed:
- get_group(): initialize g_auto(GStrv) keys with NULL

src/config-file.c
```diff
@@ -130,20 +130,43 @@ static gboolean get_key_bool(GKeyFile *key_file, const gchar *group,
...
+ * @brief Get integer value from key_file for key in group, default_value must
+ * be specified, returned in case key not found in group.
+ *
+ * @param[in]  key_file      GKeyFile to look value up
+ * @param[in]  group         A group name
+ * @param[in]  key           A key
+ * @param[out] value         Output integer value
+ * @param[in]  default_value Return this value in case no value found
+ * @param[out] error         Error
+ * @return FALSE on error (not found is no error; error is set), TRUE otherwise
```
ejoerns, line 143: a bit misleading having both statements in the same bracket, but I think I got what you meant

src/config-file.c
```diff
@@ -171,37 +171,54 @@ static gboolean get_key_int(GKeyFile *key_file, const gchar *group,
...
+        g_autoptr(GHashTable) tmp_hash = NULL;
         guint key;
         gsize num_keys;
-        gchar **keys;
+        g_auto(GStrv) keys = NULL;
+
+        g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+        g_return_val_if_fail(key_file, FALSE);
+        g_return_val_if_fail(group, FALSE);
+        g_return_val_if_fail(hash, FALSE);
```
ejoerns, line 195: should be

``` suggestion
g_return_val_if_fail(hash && *hash == NULL, FALSE);
```

src/config-file.c
```diff
@@ -222,28 +222,38 @@ static gboolean get_group(GKeyFile *key_file, const gchar *group,
...
                        G_LOG_LEVEL_INFO;
-        } else if (g_strcmp0(log_level, "debug") == 0) {
-                return G_LOG_LEVEL_MASK;
-        } else {
+        else if (g_strcmp0(log_level, "debug") == 0)
+                return G_LOG_LEVEL_ERROR | G_LOG_LEVEL_CRITICAL |
+                       G_LOG_LEVEL_WARNING | G_LOG_LEVEL_MESSAGE |
+                       G_LOG_LEVEL_INFO | G_LOG_LEVEL_DEBUG;
+        else
                 return G_LOG_LEVEL_ERROR | G_LOG_LEVEL_CRITICAL |
```
ejoerns, line 255: Maybe this is a case for a `g_error()` or at least `g_warning`. If we passed an invalid 'log_level' we should not silently accept this as this is a coding error.

src/config-file.c
```diff
@@ -256,73 +256,91 @@ static GLogLevelFlags log_level_from_string(const gchar *log_level)
...
+                                                  &config->gateway_token, NULL,
+                                                  NULL);
         if (!key_auth_token_exists && !key_gateway_token_exists) {
-                g_set_error(error, 1, 4, "Neither auth_token nor gateway_token is set in the config.");
+                g_set_error(error, G_KEY_FILE_ERROR,
+                            G_KEY_FILE_ERROR_INVALID_VALUE,
+                            "Neither auth_token nor gateway_token is set in the config.");
+                return NULL;
+        }
+        if (key_auth_token_exists && key_gateway_token_exists) {
```
ejoerns, line 295: This changes behavior, thus please briefly describe that and why this should be an error instead in the commit message.

ejoerns (changes requested):
Added some comments inside.

The mixing of line break and white space cleanups with actual changes makes it harder to figure out what was actually changed. I'd generally prefer to keep this separate as the commits already sufficiently accumulate other only partly related changes. At least for the next series.

Bastian-Krause:
> The mixing of line break and white space cleanups with actual changes makes it harder to figure out what was actually changed. I'd generally prefer to keep this separate as the commits already sufficiently accumulate other only partly related changes. At least for the next series.

Yes, you are right. I'll try to separate style changes in future PRs.

src/config-file.c
```diff
@@ -36,22 +36,48 @@ static const gboolean DEFAULT_SSL         = TRUE;
 static const gboolean DEFAULT_SSL_VERIFY  = TRUE;
 static const gchar * DEFAULT_LOG_LEVEL    = "message";
 
-static gboolean get_key_string(GKeyFile *key_file, const gchar* group, const gchar* key, gchar** value, const gchar* default_value, GError **error)
+/**
+ * @brief Get string value from key_file for key in group, optionally
+ * default_value can be specified in case key not found in group.
```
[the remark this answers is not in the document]
Bastian-Krause: Fixed, thanks.

src/config-file.c
```diff
@@ -36,22 +36,48 @@ static const gboolean DEFAULT_SSL         = TRUE;
...
                         *value = g_strdup(default_value);
+                        g_clear_error(error);
                         return TRUE;
                 }
 
-                g_set_error(error, G_KEY_FILE_ERROR,
-                            G_KEY_FILE_ERROR_NOT_FOUND,
-                            "Key '%s' not found in group '%s' and no default given", key, group);
+                g_prefix_error(error,
+                               "Key '%s' not found in group '%s' and no default given: ",
```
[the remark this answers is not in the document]
Bastian-Krause: True, removed the whole g_prefix_error() statement.

src/config-file.c
Bastian-Krause: Agreed. I improved the doc string.

src/config-file.c
Bastian-Krause: Right, thanks. Fixed.

src/config-file.c
Bastian-Krause: The value comes from the config file, so it is a configuration error. I've added a `g_warning`.

src/config-file.c
Bastian-Krause: I've moved this change into a dedicated commit with an appropriate commit message.

ejoerns (approved):
Still some `g_return_val_if_fail(error...` lines ordered wrongly, but otherwise ok now :+1:

Bastian-Krause:
> Still some `g_return_val_if_fail(error...` lines ordered wrongly, but otherwise ok now

Sorry, I somehow missed those. Fixed now.

src/config-file.c
```diff
@@ -34,53 +34,86 @@ static const gint DEFAULT_TIMEOUT         = 60;     // 1 min.
...
-{
-        gchar *val = NULL;
-        val = g_key_file_get_string(key_file, group, key, NULL);
-        if (val == NULL) {
-                if (default_value != NULL) {
+        g_return_val_if_fail(key_file, FALSE);
+        g_return_val_if_fail(group, FALSE);
+        g_return_val_if_fail(key, FALSE);
+        g_return_val_if_fail(value, FALSE);
+        g_return_val_if_fail(value && *value == NULL, FALSE);
```
ejoerns, line 61: just saw that this line now duplicates the check from the line above, so please drop the original check line.

Bastian-Krause: Fixed.

ejoerns: :+1:

PR #67, opened 2020-12-08
log: update/move doc strings

Bastian-Krause:
- add a missing return in log_level_to_string()'s doc string
- move non-static setup_logging()'s doc string to header file


ejoerns (approved):
:+1:

PR #70, opened 2021-01-05
.github: tests: use secrets for docker login only for master builds

ejoerns:
Since PR builds from forks do not have access to the GitHub repositories
secrets, a PR will fail logging in to docker:

```
Run docker/login-action@v1
  with:
    logout: true
Error: Username and password required
```

Fix this by limiting the use of login-action@v1 to master branch.


Bastian-Krause (approved):
Thank you for the quick fix!

PR #74, opened 2021-01-27
Starting point for sphinx-based user documentation

ejoerns:
This adds a basic skeleton for rauc-hawkbit-updater user documentation based on sphinx.

The tests are extended to test for doc (sphinx + doxygen) building, too.

Note that this also contains a fix for the current CI that fails to find changed debian repos.

The next step would then be to register a readthedocs subdomain for this, to allow browsing the documentation online.

PR #75, opened 2021-02-02
.github: tests: run apt-get update before install

ejoerns:
Otherwise, we could end up with a deprecated package cache and fail to install packages not downloadable anymore.


ejoerns:
I've made a separate PR for the CI fix contained in #74 to not block this by the review of the other changes included there.

Bastian-Krause (approved):
Very good, I was about to suggest exactly that :)

PR #76, opened 2021-02-17
test: create_test_target: enable target security token authentication via MGMT API

Bastian-Krause:
hawkBit's default of "authentication.targettoken.enabled" changed from true to false:

  https://github.com/eclipse/hawkbit/pull/1074

Enable the config option explicitly to make our tests work again.


ejoerns (approved):
Thanks for fixing this!

PR #77, opened 2021-02-18
hawkbit client: refactoring and bug fixes

Bastian-Krause:
This is the next PR of a series of PRs to refactor and clean up the code base, get rid of bugs (memory leaks, use-after-frees, race conditions), simplify functions and document them.

In this iteration I focused on the `hawkbit-client` module.

### DIFF
--- a/include/hawkbit-client.h
+++ b/include/hawkbit-client.h
@@ -83,14 +83,14 @@ typedef struct BinaryPayload_ {
 /**
  * @brief struct containing
  */
-struct artifact {
-        gchar* name;                  /**< name of software */
-        gchar* version;               /**< software version */
+typedef struct Artifact_ {
+        gchar *name;                  /**< name of software */
+        gchar *version;               /**< software version */
         gint64 size;                  /**< size of software bundle file */
-        gchar* download_url;          /**< download URL of software bundle file */
-        gchar* feedback_url;          /**< URL status feedback should be sent to */
-        gchar* sha1;                  /**< sha1 checksum of software bundle file */
-};
+        gchar *download_url;          /**< download URL of software bundle file */
+        gchar *feedback_url;          /**< URL status feedback should be sent to */
+        gchar *sha1;                  /**< sha1 checksum of software bundle file */
+} Artifact;
 
 /**
  * @brief struct containing the new downloaded file.
@@ -142,7 +142,15 @@ void rest_payload_free(RestPayload *payload);
  */
 void binary_payload_free(BinaryPayload *payload);
 
+/**
+ * @brief Frees the memory allocated by an Artifact
+ *
+ * @param[in] artifact Artifact to free
+ */
+void artifact_free(Artifact *artifact);
+
 G_DEFINE_AUTOPTR_CLEANUP_FUNC(RestPayload, rest_payload_free)
 G_DEFINE_AUTOPTR_CLEANUP_FUNC(BinaryPayload, binary_payload_free)
+G_DEFINE_AUTOPTR_CLEANUP_FUNC(Artifact, artifact_free)
 
 #endif // __HAWKBIT_CLIENT_H__
--- a/include/hawkbit-client.h
+++ b/include/hawkbit-client.h
@@ -73,12 +73,12 @@ typedef struct RestPayload_ {
 /**
  * @brief struct containing Curl write callback context.
  */
-struct get_binary {
+typedef struct BinaryPayload_ {
         FILE *fp;                     /**< filepointer to download file */
         gint64 filesize;              /**< expected file size of download file */
         gint64 written;               /**< number of bytes written to download file */
         GChecksum *checksum;          /**< checksum of download file */
-};
+} BinaryPayload;
 
 /**
  * @brief struct containing
@@ -120,6 +120,14 @@ gboolean install_complete_cb(gpointer ptr);
  */
 void rest_payload_free(RestPayload *payload);
 
+/**
+ * @brief Frees the memory allocated by a BinaryPayload
+ *
+ * @param[in] payload BinaryPayload to free
+ */
+void binary_payload_free(BinaryPayload *payload);
+
 G_DEFINE_AUTOPTR_CLEANUP_FUNC(RestPayload, rest_payload_free)
+G_DEFINE_AUTOPTR_CLEANUP_FUNC(BinaryPayload, binary_payload_free)
 
 #endif // __HAWKBIT_CLIENT_H__
--- a/include/hawkbit-client.h
+++ b/include/hawkbit-client.h
@@ -110,6 +110,12 @@ struct on_install_complete_userdata {
 };
 
 void hawkbit_init(Config *config, GSourceFunc on_install_ready);
+
+/**
+ * @brief Sets up timeout and event sources, initializes and runs main loop.
+ *
+ * @return numeric return code, to be returned by main()
+ */
 int hawkbit_start_service_sync();
 
 /**
--- a/include/hawkbit-client.h
+++ b/include/hawkbit-client.h
@@ -109,6 +109,14 @@ struct on_install_complete_userdata {
         gboolean install_success;               /**< status of installation */
 };
 
+/**
+ * @brief Pass config, callback for installation ready and initialize libcurl.
+ *        Intended to be called from program's main().
+ *
+ * @param[in] config Config* to make global
+ * @param[in] on_install_ready GSourceFunc to call after artifact download, to
+ *                             trigger RAUC installation
+ */
 void hawkbit_init(Config *config, GSourceFunc on_install_ready);
 
 /**
--- a/include/hawkbit-client.h
+++ b/include/hawkbit-client.h
@@ -110,6 +110,14 @@ struct on_install_complete_userdata {
 
 void hawkbit_init(Config *config, GSourceFunc on_install_ready);
 int hawkbit_start_service_sync();
+
+/**
+ * @brief Callback for install thread, sends msg as progress feedback to
+ *        hawkBit.
+ *
+ * @param[in] msg Progress message
+ * @return G_SOURCE_REMOVE is always returned
+ */
 gboolean hawkbit_progress(const gchar *msg);
 gboolean install_complete_cb(gpointer ptr);
 
--- a/include/hawkbit-client.h
+++ b/include/hawkbit-client.h
@@ -28,6 +28,13 @@
 
 #include "config-file.h"
 
+#define RHU_HAWKBIT_CLIENT_ERROR rhu_hawkbit_client_error_quark()
+GQuark rhu_hawkbit_client_error_quark(void);
+
+typedef enum {
+        RHU_HAWKBIT_CLIENT_ERROR_ALREADY_IN_PROGRESS,
+} RHUHawkbitClientError;
+
 #define HAWKBIT_USERAGENT                 "rauc-hawkbit-c-agent/1.0"
 #define DEFAULT_CURL_REQUEST_BUFFER_SIZE  512
 #define DEFAULT_CURL_DOWNLOAD_BUFFER_SIZE 64 * 1024 // 64KB
--- a/include/hawkbit-client.h
+++ b/include/hawkbit-client.h
@@ -119,6 +119,13 @@ int hawkbit_start_service_sync();
  * @return G_SOURCE_REMOVE is always returned
  */
 gboolean hawkbit_progress(const gchar *msg);
+
+/**
+ * @brief Callback for install thread, sends installation feedback to hawkBit.
+ *
+ * @param[in] ptr on_install_complete_userdata* containing set install_success
+ * @return G_SOURCE_REMOVE is always returned
+ */
 gboolean install_complete_cb(gpointer ptr);
 
 /**
--- a/include/hawkbit-client.h
+++ b/include/hawkbit-client.h
@@ -35,6 +35,14 @@ typedef enum {
         RHU_HAWKBIT_CLIENT_ERROR_ALREADY_IN_PROGRESS,
 } RHUHawkbitClientError;
 
+// uses CURLcode as error codes
+#define RHU_HAWKBIT_CLIENT_CURL_ERROR rhu_hawkbit_client_curl_error_quark()
+GQuark rhu_hawkbit_client_curl_error_quark(void);
+
+// uses HTTP codes as error codes
+#define RHU_HAWKBIT_CLIENT_HTTP_ERROR rhu_hawkbit_client_http_error_quark()
+GQuark rhu_hawkbit_client_http_error_quark(void);
+
 #define HAWKBIT_USERAGENT                 "rauc-hawkbit-c-agent/1.0"
 #define DEFAULT_CURL_REQUEST_BUFFER_SIZE  512
 #define DEFAULT_CURL_DOWNLOAD_BUFFER_SIZE 64 * 1024 // 64KB
--- a/include/hawkbit-client.h
+++ b/include/hawkbit-client.h
@@ -65,10 +65,10 @@ enum HTTPMethod {
 /**
  * @brief struct containing the payload and size of REST body.
  */
-struct rest_payload {
+typedef struct RestPayload_ {
         gchar *payload;               /**< string representation of payload */
         size_t size;                  /**< size of payload */
-};
+} RestPayload;
 
 /**
  * @brief struct containing Curl write callback context.
@@ -112,5 +112,14 @@ void hawkbit_init(Config *config, GSourceFunc on_install_ready);
 int hawkbit_start_service_sync();
 gboolean hawkbit_progress(const gchar *msg);
 gboolean install_complete_cb(gpointer ptr);
+
+/**
+ * @brief Frees the memory allocated by a RestPayload
+ *
+ * @param[in] payload RestPayload to free
+ */
+void rest_payload_free(RestPayload *payload);
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC(RestPayload, rest_payload_free)
 
 #endif // __HAWKBIT_CLIENT_H__
--- a/include/hawkbit-client.h
+++ b/include/hawkbit-client.h
@@ -33,6 +33,7 @@ GQuark rhu_hawkbit_client_error_quark(void);
 
 typedef enum {
         RHU_HAWKBIT_CLIENT_ERROR_ALREADY_IN_PROGRESS,
+        RHU_HAWKBIT_CLIENT_ERROR_JSON_RESPONSE_PARSE,
 } RHUHawkbitClientError;
 
 // uses CURLcode as error codes
--- a/include/hawkbit-client.h
+++ b/include/hawkbit-client.h
@@ -64,6 +64,14 @@ enum HTTPMethod {
 };
 
 /**
+ * @brief struct that contains the context of an HawkBit action.
+ */
+struct HawkbitAction {
+        gchar *id;                    /**< HawkBit action id */
+        GMutex mutex;                 /**< mutex used for accessing all other members */
+};
+
+/**
  * @brief struct containing the payload and size of REST body.
  */
 typedef struct RestPayload_ {
--- a/include/hawkbit-client.h
+++ b/include/hawkbit-client.h
@@ -34,6 +34,7 @@ GQuark rhu_hawkbit_client_error_quark(void);
 typedef enum {
         RHU_HAWKBIT_CLIENT_ERROR_ALREADY_IN_PROGRESS,
         RHU_HAWKBIT_CLIENT_ERROR_JSON_RESPONSE_PARSE,
+        RHU_HAWKBIT_CLIENT_ERROR_DOWNLOAD,
 } RHUHawkbitClientError;
 
 // uses CURLcode as error codes
--- a/src/hawkbit-client.c
+++ b/src/hawkbit-client.c
@@ -601,10 +601,18 @@ static gchar* build_api_url(const gchar *path, ...)
 gboolean hawkbit_progress(const gchar *msg)
 {
         g_autofree gchar *feedback_url = NULL;
-        if (action_id) {
-                feedback_url = build_api_url("deploymentBase/%s/feedback", action_id);
-                feedback_progress(feedback_url, action_id, msg, NULL);
-        }
+        g_autoptr(GError) error = NULL;
+
+        g_return_val_if_fail(msg, FALSE);
+
+        if (!action_id)
+                return G_SOURCE_REMOVE;
+
+        feedback_url = build_api_url("deploymentBase/%s/feedback", action_id);
+
+        if (!feedback_progress(feedback_url, action_id, msg, &error))
+                g_warning("%s", error->message);
+
         return G_SOURCE_REMOVE;
 }
 
--- a/src/hawkbit-client.c
+++ b/src/hawkbit-client.c
@@ -68,7 +68,7 @@ static const char *HTTPMethod_STRING[] = {
 
 static Config *hawkbit_config = NULL;
 static GSourceFunc software_ready_cb;
-static gchar * volatile action_id = NULL;
+static struct HawkbitAction *active_action = NULL;
 static GThread *thread_download = NULL;
 
 GQuark rhu_hawkbit_client_error_quark(void)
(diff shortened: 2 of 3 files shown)


Bastian-Krause:
Note that the failing authentication test is caused by a change in hawkBit, resolved in #76.

Bastian-Krause:
Force-pushed: rebased on current master.

src/hawkbit-client.c
```diff
@@ -135,7 +145,7 @@ static gboolean get_binary(const gchar* download_url, const gchar* file, gint64
         CURL *curl = curl_easy_init();
         if (!curl) {
                 fclose(fp);
-                g_set_error(error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                g_set_error(error, RHU_HAWKBIT_CLIENT_CURL_ERROR, -1,
```
ejoerns, line 265: The icing on the cake would be to define an error code for this like `RHU_HAWKBIT_CLIENT_ERROR_INIT`.

src/hawkbit-client.c
```diff
@@ -231,9 +250,12 @@ static gint rest_request(enum HTTPMethod method, const gchar* url, JsonBuilder*
 {
         gchar *postdata = NULL;
         struct rest_payload fetch_buffer;
-
+        glong http_code = 0;
         CURL *curl = curl_easy_init();
-        if (!curl) return -1;
+        if (!curl) {
+                g_set_error(error, RHU_HAWKBIT_CLIENT_CURL_ERROR, -1, "curl init failed");
```
ejoerns, line 374: The same above produces the message `"Unable to start libcurl easy session"`. We should make that consistent in any direction.

src/hawkbit-client.c
```diff
@@ -303,20 +324,14 @@ static gint rest_request(enum HTTPMethod method, const gchar* url, JsonBuilder*
...
-                            http_code,
-                            "HTTP request timed out: %s",
-                            curl_easy_strerror(res));
-        } else {
-                g_set_error(error,
-                            1,                    // error domain
-                            http_code,
-                            "HTTP request failed: %s",
+        }
+        if (res != CURLE_OK) {
```
ejoerns, line 420: When `res` becomes FALSE, so we have any guarantee on http_code? Otherwise funtions will still return error indicator and set GError incosistently.

include/hawkbit-client.h
```diff
@@ -113,4 +113,13 @@ int hawkbit_start_service_sync();
...
 gboolean install_complete_cb(gpointer ptr);
 
+/**
+ * @brief Frees the memory allocated by a RestPayload
+ *
+ * @param[in] payload RestPayload to free
+ */
+void rest_payload_free(RestPayload *payload);
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC(RestPayload, rest_payload_free)
```
ejoerns, line 175: We prepare this but do not make use of it here.

src/hawkbit-client.c
```diff
@@ -249,7 +249,7 @@ static size_t curl_write_cb(void *content, size_t size, size_t nmemb, void *data
 static gint rest_request(enum HTTPMethod method, const gchar* url, JsonBuilder* jsonRequestBody, JsonParser** jsonResponseParser, GError** error)
 {
         gchar *postdata = NULL;
-        struct rest_payload fetch_buffer;
+        RestPayload *fetch_buffer = g_new0(RestPayload, 1);
```
ejoerns: maybe note that you change from stack allocation to heap allocation here

src/hawkbit-client.c
```diff
@@ -121,6 +121,34 @@ static size_t curl_write_to_file_cb(void *ptr, size_t size, size_t nmemb, struct
...
+ * @param[out] headers curl_slist** of already set headers
+ * @param[in]  string  Header to add
+ * @param[out] error   Error
+ * @return TRUE if string was added to headers successfully, FALSE otherwise (error set)
+ */
+static gboolean add_curl_header(struct curl_slist **headers, const char *string, GError **error)
+{
+        struct curl_slist *temp = NULL;
+
+        g_return_val_if_fail(string, FALSE);
```
ejoerns, line 169: should also check for `curl_slist` being `NULL` as dereferencing below will fail otherwise.

src/hawkbit-client.c
```diff
@@ -121,6 +121,34 @@ static size_t curl_write_to_file_cb(void *ptr, size_t size, size_t nmemb, struct
...
+        g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+        temp = curl_slist_append(*headers, string);
+        if (!temp) {
+                curl_slist_free_all(*headers);
+                g_set_error(error, RHU_HAWKBIT_CLIENT_CURL_ERROR, -1, "Could not add header %s",
```
ejoerns, line 175: Should we use -1 here, too?

src/hawkbit-client.c
```diff
@@ -272,27 +272,29 @@ static size_t curl_write_cb(void *content, size_t size, size_t nmemb, void *data
...
         g_autofree gchar *token = NULL;
         RestPayload *fetch_buffer = g_new0(RestPayload, 1);
         glong http_code = 0;
         struct curl_slist *headers = NULL;
         CURL *curl = curl_easy_init();
+        gboolean ret = TRUE;
         if (!curl) {
                 g_set_error(error, RHU_HAWKBIT_CLIENT_CURL_ERROR, -1, "curl init failed");
-                return -1;
+                ret = FALSE;
```
ejoerns: The actual changes in this PR are mainly fixing the cleanup handling. But this is not what the commit messages states.
Thus either split this up or describe this explicitly.

src/hawkbit-client.c
```diff
@@ -333,47 +337,55 @@ static gint rest_request(enum HTTPMethod method, const gchar* url, JsonBuilder*
...
+                goto clean;
         } else if (http_code != 200) {
                 g_set_error(error, RHU_HAWKBIT_CLIENT_HTTP_ERROR, http_code,
                             "HTTP request failed: %ld; server response: %s", http_code,
                             fetch_buffer->payload);
+                ret = FALSE;
+                goto clean;
         }
 
+        if (jsonResponseParser && fetch_buffer->size > 0) {
```
ejoerns, line 432: this change seems to belong logically to the rework of error code handling some commits ago, not to return value or cleanup handling.

src/hawkbit-client.c
```diff
@@ -265,40 +267,42 @@ static size_t curl_write_cb(void *content, size_t size, size_t nmemb, void *data
...
+        g_return_val_if_fail(url, FALSE);
+        g_return_val_if_fail(jsonResponseParser == NULL || *jsonResponseParser == NULL, FALSE);
+        g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+        curl = curl_easy_init();
         if (!curl) {
                 g_set_error(error, RHU_HAWKBIT_CLIENT_CURL_ERROR, -1, "curl init failed");
-                ret = FALSE;
-                goto clean;
+                return FALSE;
```
ejoerns, line 375: This actually reverts flow changes made in the commit before, maybe doing the autocleanup support earlier will a bunch of unnecessary changes?

src/hawkbit-client.c
```diff
@@ -245,17 +245,24 @@ static gboolean get_binary(const gchar* download_url, const gchar* file, gint64
...
+
+        p = (RestPayload *) data;
+
         p->payload = (gchar *) g_realloc(p->payload, p->size + real_size + 1);
-        if (p->payload == NULL) {
-                g_debug("Failed to expand buffer");
-                return -1;
+        if (!p->payload) {
+                g_error("Failed to expand buffer");
+                return 0;
```
ejoerns, line 336: return after g_error will likely never be called but abort the program.

ejoerns (changes requested):
I've reviewed the first part of the PR, but interrupted here as future commits might depend too much on change requests already noted. Please review/fix these first.

I also have to note that this is quite a lot for a single PR which does not necessarily make things easier :smirk:

src/hawkbit-client.c
```diff
@@ -589,7 +595,10 @@ static gboolean process_deployment(JsonNode *req_root, GError **error)
         struct artifact *artifact = NULL;
 
         if (action_id) {
-                g_warning("Deployment is already in progress...");
+                g_set_error(error, RHU_HAWKBIT_CLIENT_ERROR,
```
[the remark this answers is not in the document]
Bastian-Krause: Right, will be mentioned in the respective commit message.

src/hawkbit-client.c
```diff
@@ -180,19 +190,28 @@ static gboolean get_binary(const gchar* download_url, const gchar* file, gint64
...
-        } else {
-                g_set_error(error,
-                            G_IO_ERROR,                    // error domain
-                            G_IO_ERROR_FAILED,             // error code
-                            "HTTP request failed: %s (%ld)",     // error message format string
-                            curl_easy_strerror(res),
-                            http_code);
+        }
+
+        if (res != CURLE_OK) {
```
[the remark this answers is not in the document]
Bastian-Krause: I did not change the meaning of `res`, did I? `res` existed before this PR as it does now.

src/hawkbit-client.c
```diff
@@ -180,19 +190,28 @@ static gboolean get_binary(const gchar* download_url, const gchar* file, gint64
...
+                            curl_easy_strerror(res));
+                ret = FALSE;
+        }
+        if (http_code != 200) {
+                g_set_error(error, RHU_HAWKBIT_CLIENT_HTTP_ERROR, http_code,
+                            "HTTP request failed: %ld", http_code);
+                ret = FALSE;
+        }
+
+        if (ret && gb.checksum) { // if checksum enabled then return the value
```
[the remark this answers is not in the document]
Bastian-Krause: True. I'll drop the redundant setting of `sha1sum`.

src/hawkbit-client.c
Bastian-Krause: Hm, this error occurs only if `curl_easy_init()` returned NULL. Isn't that a curl error?

src/hawkbit-client.c
Bastian-Krause: You're right. I'll use "Unable to start libcurl easy session" for both cases.

src/hawkbit-client.c
Bastian-Krause: Let's use `CURLE_FAILED_INIT` for all those `-1` cases. Do you want to be able to distinguish these cases any further? I think that's not needed.

src/hawkbit-client.c
Bastian-Krause: Right, changing the return type and refactoring of `rest_request()` got mixed up here. Will fix it.

src/hawkbit-client.c
Bastian-Krause: Yep, changing the return type and refactoring of `rest_request()` got mixed up, so the whole intermediate flow changes can be dropped. I'll do that.

src/hawkbit-client.c
Bastian-Krause: Right, changing the return type and refactoring of rest_request() got mixed up. Will fix it.

src/hawkbit-client.c
Bastian-Krause: I guess it is okay that a failing `g_realloc()` aborts. We can't recover from that. The return statement is there for the sake of understandability.

src/hawkbit-client.c
Bastian-Krause: I don't understand this. In case the curl operation does not succeed (`!= CURLE_OK`), the appropriate RHU_HAWKBIT_CLIENT_CURL_ERROR will be set and the functions returns FALSE. Only if `res` is `CURLE_OK` the `http_code` will be checked. Or am I missing something here?

include/hawkbit-client.h
Bastian-Krause: That's correct. I think that defining the cleanup function to be used for RestPayload g_autoptr here is cleaner compared to introducing this in the first function refactoring commit where it is actually used.

Bastian-Krause:
I started to incorporate the feedback. I'll continue with that tomorrow and push the changes.

If you want me to split this PR into multiple parts, please make a suggestion where to split the patch stack. I cannot find a logical point to split it up..

src/hawkbit-client.c
ejoerns: I was referring to something like

``` c
#define  RHU_HAWKBIT_CLIENT_CURL_ERROR_INIT -1
```

and

```suggestion
                g_set_error(error, RHU_HAWKBIT_CLIENT_CURL_ERROR, RHU_HAWKBIT_CLIENT_CURL_ERROR_INIT,
```

src/hawkbit-client.c
ejoerns: You do not neet to check for the return actually as both the function description as well as the implementation do not state to signalize errors with returning NULL.
Instead, the function will fail itself with g_error() when allocatio fails. Refer: https://gitlab.gnome.org/GNOME/glib/-/blob/master/glib/gmem.c#L176

src/hawkbit-client.c
Bastian-Krause: Done.

src/hawkbit-client.c
Bastian-Krause: Done.

src/hawkbit-client.c
Bastian-Krause: Right, dropped the `if` altogether

src/hawkbit-client.c
Bastian-Krause: I see, I've used `CURLE_FAILED_INIT` now instead.

Bastian-Krause:
> I started to incorporate the feedback. I'll continue with that tomorrow and push the changes.
Reality kicked in, so it took longer than expected to address the feedback. So finally, here it is.

src/hawkbit-client.c
ejoerns: Sounds sufficient, yes.

src/hawkbit-client.c
ejoerns: Ok, this behavior is changed in a later commit. When writing this note the method did still say `return http_code`.
(refer https://github.com/rauc/rauc-hawkbit-updater/pull/77/commits/a57e14ad4ed3878b809563907059d8a7e24dde2c)
For this commit that only claims to "introduce curl error domain", it should be fine.

ejoerns (changes requested):
Some minor things left to be fixed before we can finally merge this impressive PR.

src/hawkbit-client.c
```diff
@@ -260,39 +262,43 @@ static size_t curl_write_cb(void *content, size_t size, size_t nmemb, void *data
...
- * @brief Make REST request.
+ * @brief Perform REST request with JSON data, expecting response JSON data.
  *
- * @param[in]  method             HTTP Method ex. GET
+ * @param[in]  method             HTTP Method, e.g. GET
  * @param[in]  url                URL used in HTTP REST request
- * @param[in]  jsonRequestBody    REST request body. If NULL, no body is sent.
- * @param[out] jsonResponseParser REST response
+ * @param[in]  jsonRequestBody    REST request body. If NULL, no body is sent
+ * @param[out] jsonResponseParser REST response. If NULL, response is not parsed
```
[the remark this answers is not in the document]
Bastian-Krause: Agreed.

src/hawkbit-client.c
```diff
@@ -487,16 +487,39 @@ static JsonBuilder* json_build_status(const gchar *id, const gchar *detail, cons
 
 /**
  * @brief Send feedback to hawkBit.
+ *
+ * @param[in]  url        hawkBit URL used for request
+ * @param[in]  id         hawkBit action ID or NULL (configData usecase)
```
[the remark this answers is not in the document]
Bastian-Krause: Oh right, that's a copy/paste error (copied from `json_build_status()`). The action ID must not be NULL here. Will fix the docstring.

src/hawkbit-client.c
```diff
@@ -487,16 +487,39 @@ static JsonBuilder* json_build_status(const gchar *id, const gchar *detail, cons
...
+ * @param[in]  detail     Detail message or NULL (configData usecase)
```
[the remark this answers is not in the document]
Bastian-Krause: Same answer as above, will fix the docstring.

src/hawkbit-client.c
```diff
@@ -213,131 +342,142 @@ static size_t curl_write_cb(void *content, size_t size, size_t nmemb, void *data
...
- * @brief Build JSON status request.
+ * @brief Build hawkBit JSON request.
+ *
  * @see https://www.eclipse.org/hawkbit/rest-api/rootcontroller-api-guide/#_post_tenant_controller_v1_controllerid_deploymentbase_actionid_feedback
+ *
+ * @param[in] id         hawkBit action ID or NULL (configData usecase)
+ * @param[in] detail     Detail message or NULL (configData usecase)
+ * @param[in] finished   hawkBit status of the result
+ * @param[in] execution  hawkBit status of the action execution
+ * @param[in] attributes hawkBit controller attributes
```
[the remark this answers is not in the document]
Bastian-Krause: Correct, see above.

src/hawkbit-client.c
```diff
@@ -213,131 +342,142 @@ static size_t curl_write_cb(void *content, size_t size, size_t nmemb, void *data
...
  */
-static void json_build_status(JsonBuilder *builder, const gchar *id, const gchar *detail, const gchar *result, const gchar *execution, GHashTable *data)
+static JsonBuilder* json_build_status(const gchar *id, const gchar *detail, const gchar *finished,
+                                      const gchar *execution, GHashTable *attributes)
 {
         GHashTableIter iter;
         gpointer key, value;
-
-        // Get current time in UTC
+        g_autoptr(JsonBuilder) builder = NULL;
```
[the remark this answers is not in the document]
Bastian-Krause: As the critical log message emitted by `g_return_val_if_fail()` is not fatal yet, we could return there. When the critical message is turned fatal, you're right, it is not strictly required. Any objection against leaving it this way though?

src/hawkbit-client.c
```diff
@@ -369,67 +509,106 @@ static void json_build_status(JsonBuilder *builder, const gchar *id, const gchar
...
-        long poll_sleep_time;
 
-        sleeptime_str = json_get_string(root, "$.config.polling.sleep", NULL);
+        g_return_val_if_fail(root, hawkbit_config->retry_wait);
+
+        sleeptime_str = json_get_string(root, "$.config.polling.sleep", &error);
         if (!sleeptime_str) {
-                poll_sleep_time = hawkbit_config->retry_wait;
-                goto out;
+                g_warning("Polling sleep time not found: %s", error->message);
```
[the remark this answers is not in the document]
Bastian-Krause: I like that, will use it :)

src/hawkbit-client.c
```diff
@@ -668,12 +684,16 @@ static gboolean identify(GError **error)
...
  * bundle at config's bundle_download_location.
+ * Must be called under locked active_action->mutex.
  */
 static void process_deployment_cleanup()
 {
-        gpointer ptr = action_id;
+        gpointer ptr;
 
-        action_id = NULL;
+        g_return_if_fail(!g_mutex_trylock(&active_action->mutex));
```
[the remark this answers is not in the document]
Bastian-Krause: True, I'll do the same in `process_deployment()`.

src/hawkbit-client.c
```diff
@@ -462,247 +641,308 @@ static gchar* build_api_url(const gchar *path, ...)
...
 static void process_deployment_cleanup()
 {
-        //g_clear_pointer(action_id, g_free);
-        gpointer ptr = action_id;
-        action_id = NULL;
+        gpointer ptr;
+
+        g_return_if_fail(!g_mutex_trylock(&active_action->mutex));
+
+        ptr =  active_action->id;
```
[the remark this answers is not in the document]
Bastian-Krause: Right, will drop that.

src/hawkbit-client.c
```diff
@@ -369,67 +509,106 @@ static void json_build_status(JsonBuilder *builder, const gchar *id, const gchar
...
 static long json_get_sleeptime(JsonNode *root)
 {
-        gchar *sleeptime_str = NULL;
+        g_autofree gchar *sleeptime_str = NULL;
+        g_autoptr(GError) error = NULL;
         struct tm time;
-        long poll_sleep_time;
 
-        sleeptime_str = json_get_string(root, "$.config.polling.sleep", NULL);
+        g_return_val_if_fail(root, hawkbit_config->retry_wait);
```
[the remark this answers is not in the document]
Bastian-Krause: Right, the sleep time is always expected. So if it is not there, that's an error. Will use your suggestion.

src/hawkbit-client.c
```diff
@@ -462,247 +641,308 @@ static gchar* build_api_url(const gchar *path, ...)
...
-        if (g_strcmp0(artifact->sha1, sha1sum)) {
-                g_autofree gchar *msg = g_strdup_printf(
-                        "Software: %s V%s. Invalid checksum: %s expected %s",
-                        artifact->name, artifact->version,
-                        sha1sum,
-                        artifact->sha1);
-                feedback(artifact->feedback_url, action_id, msg, "failure", "closed", NULL);
-                g_critical("%s", msg);
-                goto down_error;
+        res = !g_strcmp0(artifact->sha1, sha1sum);
```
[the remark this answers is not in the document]
Bastian-Krause: Okay.

src/hawkbit-client.c
```diff
@@ -462,247 +641,308 @@ static gchar* build_api_url(const gchar *path, ...)
...
-        // setup checksum
-        g_autofree gchar *sha1sum = NULL;
+        g_return_val_if_fail(data, NULL);
+
+        g_message("Start downloading: %s", artifact->download_url);
 
         // Download software bundle (artifact)
-        gboolean res = get_binary(artifact->download_url, hawkbit_config->bundle_download_location,
-                                  artifact->size, &sha1sum, &speed, &error);
+        res = get_binary(artifact->download_url, hawkbit_config->bundle_download_location,
```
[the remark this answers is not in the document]
Bastian-Krause: Right. Will drop it.

src/hawkbit-client.c
```diff
@@ -462,247 +641,308 @@ static gchar* build_api_url(const gchar *path, ...)
...
+        goffset freespace;
+
+        g_return_val_if_fail(req_root, FALSE);
+        g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+        g_return_val_if_fail(!g_mutex_trylock(&active_action->mutex), FALSE);
+
+        if (active_action->id) {
+                g_set_error(error, RHU_HAWKBIT_CLIENT_ERROR,
+                            RHU_HAWKBIT_CLIENT_ERROR_ALREADY_IN_PROGRESS,
+                            "Deployment %s is already in progress.", active_action->id);
```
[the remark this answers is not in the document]
Bastian-Krause: Thanks for mentioning this, that was not intended. Will fix it.

src/hawkbit-client.c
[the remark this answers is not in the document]
ejoerns: It's ok for me to leave it.

Bastian-Krause:
Force-pushed: addressed @ejoerns' feedback. Thanks!

src/hawkbit-client.c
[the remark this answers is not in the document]
ejoerns: But you did not apply the suggestion, right?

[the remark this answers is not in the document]
Bastian-Krause: Oops, fixed and pushed.

src/hawkbit-client.c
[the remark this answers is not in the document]
Bastian-Krause: I answered with
> Both the JsonNode and the sleep time are always expected. So if these are not there, that's an error. I'll return 0 for both cases.

Now I'm unsure about this. Should we handle such a case (which should never happen according to the DDI API documentation) with a warning and carry on with `hawkbit_config->retry_wait`?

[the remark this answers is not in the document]
ejoerns: > I answered with

Ok, was lost in review comments.

I'd say it is okay to abort if the API breaks unexpectedly, but then I would semantically not use `g_return_val_if_fail()` but `g_error()` or similar instead.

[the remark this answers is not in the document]
Bastian-Krause: I could add a `GError**` argument to `json_get_sleeptime()`. Looking at the function calling `json_get_sleeptime()` (being `hawkbit_pull_cb()`), this would probably lead to an immediate..
```
g_warning("%s", error->message);
g_clear_error(&error);
```
Then we would wait for either the last sleep time we got or `hawkbit_config->retry_wait`.

I don't know if that's worth the hassle. Maybe we should go back to the initial way and simply emit a warning here directly and live with `hawkbit_config->retry_wait`?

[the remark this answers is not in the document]
ejoerns: > I don't know if that's worth the hassle. Maybe we should go back to the initial way and simply emit a warning here directly and live with hawkbit_config->retry_wait?

I'd be fine with that too, yes.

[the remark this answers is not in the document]
ejoerns: I've simply re-applied the (slightly modified) warning variant now to resolve this PR.

ejoerns (approved):

ejoerns:
Many thanks @Bastian-Krause for your effort!

Issue #73, opened 2021-01-26
Status command


SijmenHuizenga:
We are working on an embedded device where the user will have to manually trigger the download & install of an update when one is available. We will be calling `rauc-hawkbit-updater --run-once` whenever the user requests to install an update. We want to show the details of the upcoming update to the user before downloading.

We need a way to know what rauc-hawkbit-updater would do if it were invoked. I propose we add the `--status` command that prints in a single-line status message and exits with a status code. For example:

| exit code | message |
|---|---|
| 0 | Everything up-to-date. No updates available. | 
| 1 | Update available: \<name\> (\<version\>) | 

More statuses could be added.

The `--status` flag would only work when running together with `--run-once`.

What do you think? Let's talk! I am happy to implement this if we agree on the interface.

Bastian-Krause:
FYI: `--run-once` does not work for this at the moment, neither the download nor the install thread can finish in this case (c65e3c5). I have patches in my queue that are waiting for #68, #71, and a (to be created) "hawkbit client refactoring" PR to get merged before. These patches make the rauc-hawkbit-updater wait for the download and install threads properly. I needed them to allow download/install test cases (also waiting in queue).

As far as I can see hawkBit is not expecting such a scenario you desribe, you targets will be marked overdue as they do not poll regularly. Also you won't be able to gather information about your targets (see [config data](https://www.eclipse.org/hawkbit/rest-api/rootcontroller-api-guide/#_put_tenant_controller_v1_controllerid_configdata)) - at least not until the user triggers rauc-hawkbit-updater.

I'm wondering if user approval for download and/or installation couldn't be integrated in a fashion closer to what hawkBit expects, with rauc-hawkbit-updater constantly running.  This again would probably require a rauc-hawkbit-updater DBus interface.

Maybe let's take a step back: what are the advantages of using rauc-hawkbit-updater for your case compared to poll/download/install a bundle by your application from a regular HTTP server?

SijmenHuizenga:
Thanks a lot @Bastian-Krause for the details explanation :heart:. I now see the problem with using `--run-once`, so we won't be using that. The plan changed to downloading the files directly and calling rauc through the d-bus api. Thanks :smile:

PR #80, opened 2021-03-30
test preparations: make threads wait for each other on --run-once, propagate results

Bastian-Krause:
We have a main loop, which starts the download thread if an update is available for the target. The download thread starts the install thread once the download succeeded. Currently the threads do not wait for each other to finish on `-r, --run-once`, but terminate early.

In order to test download and installation cases with `-r, --run-once` and allow use cases such as #73, the threads need to wait for each other to finish. It is also required to emit an appropriate exit code for download and installation errors.

This PR allows the threads to wait for each other to finish on `-r, --run-once` and implements status propagation.

### DIFF
--- a/include/hawkbit-client.h
+++ b/include/hawkbit-client.h
@@ -108,6 +108,7 @@ struct on_new_software_userdata {
         GSourceFunc install_progress_callback;  /**< callback function to be called when new progress */
         GSourceFunc install_complete_callback;  /**< callback function to be called when installation is complete */
         gchar *file;                            /**< downloaded new software file */
+        gboolean install_success;               /**< whether the installation succeeded or not (only meaningful for run_once mode!) */
 };
 
 /**
--- a/src/hawkbit-client.c
+++ b/src/hawkbit-client.c
@@ -1040,7 +1040,7 @@ out:
                         res = GPOINTER_TO_INT(thread_ret);
                 }
 
-                data->res = res ? 0 : 1;
+                data->res = res;
                 g_main_loop_quit(data->loop);
                 return G_SOURCE_REMOVE;
         }
@@ -1097,7 +1097,7 @@ int hawkbit_start_service_sync()
 
         g_main_loop_run(cdata.loop);
 
-        res = cdata.res;
+        res = cdata.res ? 0 : 1;
 
 #ifdef WITH_SYSTEMD
         sd_notify(0, "STOPPING=1\nSTATUS=Stopped polling HawkBit for new software.");
--- a/src/hawkbit-client.c
+++ b/src/hawkbit-client.c
@@ -741,7 +741,10 @@ gboolean install_complete_cb(gpointer ptr)
  * feedback and call software_ready_cb() callback on success.
  *
  * @param[in] data Artifact* to process
- * @return NULL is always returned
+ * @return gpointer being 1 (TRUE) if download succeeded, 0 (FALSE) otherwise. The return value is
+ *         meant to be used with the GPOINTER_TO_INT() macro only.
+ *         Note that if the download thread waited for installation to finish ('run_once' mode),
+ *         TRUE means both installation and download succeeded.
  */
 static gpointer download_thread(gpointer data)
 {
@@ -796,7 +799,7 @@ static gpointer download_thread(gpointer data)
 
         software_ready_cb(&userdata);
 
-        return NULL;
+        return GINT_TO_POINTER(userdata.install_success);
 
 report_err:
         g_mutex_trylock(&active_action->mutex);
@@ -807,7 +810,7 @@ report_err:
         process_deployment_cleanup();
         g_mutex_unlock(&active_action->mutex);
 
-        return NULL;
+        return GINT_TO_POINTER(FALSE);
 }
 
 /**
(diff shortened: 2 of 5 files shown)


src/hawkbit-client.c
```diff
@@ -741,19 +742,23 @@ gboolean install_complete_cb(gpointer ptr)
  * feedback and call software_ready_cb() callback on success.
  *
  * @param[in] data Artifact* to process
- * @return NULL is always returned
+ * @return if download thread waited for installation to finish, TRUE if installation and download
+ *         succeeded, FALSE otherwise. If download thread did not wait for installation to finish,
+ *         TRUE if download succeeded, FALSE otherwise.
```
ejoerns, line 747: Maybe it's also an option to use `GINT_TO_POINTER()` here?

ejoerns, line 747:
``` suggestion 
 * @return pointer to gboolean being TRUE if download succeeded, FALSE otherwise.
 *         Note that if the download thread waited for installation to finish ('run_once' mode),
 *         TRUE means both installation and download succeeded.
```

src/hawkbit-client.c
```diff
@@ -806,7 +812,8 @@ static gpointer download_thread(gpointer data)
         process_deployment_cleanup();
         g_mutex_unlock(&active_action->mutex);
 
-        return NULL;
+        *thread_ret = FALSE;
+        return g_steal_pointer(&thread_ret);
```
ejoerns, line 816: This also means we always have to free the return of `g_thread_join()` right? Currently in the unref before (re)creating this thread we don't do that.

ejoerns (commented):
I got lost a bit when trying to untangle my brain for reviewing this, thus please check if my comments make sense for you :smirk:

src/hawkbit-client.c
```diff
@@ -795,7 +800,8 @@ static gpointer download_thread(gpointer data)
 
         software_ready_cb(&userdata);
 
-        return NULL;
+        *thread_ret = userdata.install_success;
```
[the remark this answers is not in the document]
Bastian-Krause: Sounds good, will do.

src/hawkbit-client.c
```diff
@@ -56,6 +56,7 @@
 #include "hawkbit-client.h"
 
 G_DEFINE_AUTOPTR_CLEANUP_FUNC(CURL, curl_easy_cleanup)
+G_DEFINE_AUTOPTR_CLEANUP_FUNC(gboolean, g_free)
```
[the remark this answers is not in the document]
Bastian-Krause: I'll drop this in favor of the suggested `GINT_TO_POINTER()`/`GINT_TO_POINTER()`.

include/hawkbit-client.h
```diff
@@ -108,6 +108,7 @@ struct on_new_software_userdata {
         GSourceFunc install_progress_callback;  /**< callback function to be called when new progress */
         GSourceFunc install_complete_callback;  /**< callback function to be called when installation is complete */
         gchar *file;                            /**< downloaded new software file */
+        gboolean install_success;               /**< wheter the installation succeeded or not */
```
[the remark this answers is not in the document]
Bastian-Krause: Applied with a small change.

src/hawkbit-client.c
Bastian-Krause: Yep!

Bastian-Krause: Thanks, applied.

src/hawkbit-client.c
Bastian-Krause: Please take a look at the new changes. I guess that's not needed anymore, right?

Bastian-Krause:
Force-pushed: addressed feedback by @ejoerns. Thanks for that. While at it, fixed gboolean/gint confusion in a separate commit.

src/hawkbit-client.c
```diff
@@ -741,14 +741,17 @@ gboolean install_complete_cb(gpointer ptr)
  * feedback and call software_ready_cb() callback on success.
  *
  * @param[in] data Artifact* to process
- * @return NULL is always returned
+ * @return pointer to gboolean being TRUE if download succeeded, FALSE otherwise.
```
ejoerns, line 744: With the change to `GINT_TO_POINTER()` We should be consistent here, too, and say someting like:

``` suggestion
 * @return gpointer being 1 (TRUE) if download succeeded, 0 (FALSE) otherwise. The return value is meant to be used with GPOINTER_TO_INT() macro only.
```

ejoerns (changes requested):
Looks much smarter now in my opinion. :+1:  Two final remarks left ;)

src/hawkbit-client.c
```diff
@@ -917,7 +919,7 @@ static gboolean process_deployment(JsonNode *req_root, GError **error)
 
         // unref/free previous download thread by joining it
         if (thread_download)
-                g_thread_join(thread_download);
+                g_free(g_thread_join(thread_download));
```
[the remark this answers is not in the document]
Bastian-Krause: Good catch, will be fixed.

src/hawkbit-client.c
Bastian-Krause: Applied :+1:

Bastian-Krause:
Force-pushed: integrated latest feedback from @ejoerns, thanks!

ejoerns (approved):

PR #82, opened 2021-03-31
RFC: Displaying Version Information Properly (git-based)

ejoerns:
We've had a few issues in the past already where people tried to figure out the version of rauc-hawkbit-updater they run.
Unfortunately, all git revisions will currently display:

``` shell
$ rauc-hawkbit-updater -v
Version 0.1.0
```

As this value is hard-coded. This is quite misleading as a) we do not even have a 0.1.0 release yet and b) it does not reflect the additional modifications.

Thus I propose to generate a version number based on `git describe --tags` that will emit something like `Version 0.2.0` for releases, but also shows 'intermediate' git states like `Version 0.2.0-176-gf497571`.

I'm not that deep within cmake, thus I copied an existing approach. Note this requires to re-run `cmake ..` in order to let git history changes take effect.

Also note: As we currently don't have any tag set, the output falls back to the commits git sha. Thus for this to work we should add at least a `v0` tag or similar until we make an actual release tag.

I've split the work up a bit and initially moved the version notation from the c file to the cmake project as the central definiton point, but that should only be relevant in case we decide against git-based versions.


jluebbe:
You need to be very careful these automatisms. Are you sure that an outer git repository is not picked up when building from a tarball? How is the version generated when building a tarball?

Bastian-Krause:
I like the idea. Most build systems won't probably clone the git, but rely on either a release tarball or create a tarball from a git revision/tag themselves. I did a quick test for that scenario:

```
$ rauc-hawkbit-updater -v
Version HEAD-HASH-NOTFOUND
```

With the introduced cmake magic, is there a way to set this version on compile time?

ejoerns:
Pretty sure we can combine this in any way, but first we need to get an idea on how to create release tarballs with cmake as I don't know of any autotoolish `make dist` equivalent where we could write a tarball-version file or similar.
Is 'cpack' the way to go here?.

Another option could be to combine a hard-coded version string with added git info in case it is available.

ejoerns:
Since we intend to drop cmake in favor of meson, I would close this draft.

Let's see if and how we resolve this topic there.

PR #83, opened 2021-04-06
test: add pytest-based test suite with download/installation tests

Bastian-Krause:
The rauc-hawkbit-updater code base is becoming increasingly complex, although its use cases are of a very limited nature.

In order to keep the code base in good shape and to prevent future bugs, having basic tests for the few use cases rauc-hawkbit-updater supports is very helpful. The current (cumbersome and limited) sharness test suite does only test command line arguments as well as registering a target at hawkBit, nothing more.

Having a test suite that actually tests the use cases of a real world updating scenario reduces future development cycles and time.

This PR adds a pytest test suite based on the previous tests, extends it with actual installation tests against a D-BUS SessionBus RAUC dummy (acting like RAUC) as well as a number of download tests (some using nginx' reverse proxy features). In order to control the hawkBit test instance, a hawkBit management API helper is provided (allowing all actions that are available in the hawkBit UI).

As before, the test suite also runs in our GitHub Actions workflow.

The sharness tests are finally dropped.


Bastian-Krause:
Force-pushed: rebased on current master

Bastian-Krause:
Force pushed:
- use `pexpect.spawn.terminate(force=True)` in `rauc_dbus_install_success()` (`rauc_dbus_dummy` child process does always terminate otherwise)
- assert both `pexpect.spawn.terminate()` calls (makes sure that the child process did really terminate)
- add more documentation to `rauc_dbus_dummy.py`'s `Installer` class

ejoerns (dismissed):
Had a look into this: Looks fine for me and is far more than we had before! :+1:  Good work!

Bastian-Krause:
Force-pushed:
- allow running pytest from within `test/` by using `__file__`'s location for `cwd` in `rauc_dbus_install_success`/`rauc_dbus_install_failure`
- fixed docstring of `test_download_inexistent_location`
- fixed docstring of `test_download_unallowed_location`

Bastian-Krause:
Force-pushed:
- licensed `script/hawkbit-mgmt.py` under 0BSD
- added 0BSD license test to `script/`
- moved `test/hawkbit.py` to `script/hawkbit-mgmt.py`
- added symlink `test/hawkbit-mgmt.py -> script/hawkbit-mgmt.py` to allow import in `conftest.py`

Bastian-Krause:
Force-pushed (licensing is hard):
- moved `script/LICENSE-0BSD -> script/LICENSE.0BSD`
- moved copyright lines from `script/LICENSE.0BSD` to `script/hawkbit_mgmt.py` (as `SPDX-FileCopyrightText`)

Bastian-Krause:
Force-pushed:
- added SPDX-License-Identifier to all .py files

test/conftest.py
```diff
@@ -0,0 +1,260 @@
+# SPDX-License-Identifier: LGPL-2.1
```
ejoerns, line 1: The rest of the code uses "LGPL-2.1-only" as license identifier. Maybe it is better to keep this consistent.
Also note that this one (compared to "LGPL-2.1") also does not have a deprecation warning on spdx.org: https://spdx.org/licenses/LGPL-2.1-only.html vs https://spdx.org/licenses/LGPL-2.1.html (despite both actually state to be "LGPL-2.1 only".. )

Bastian-Krause:
Force-pushed:
- use license identifier "LGPL-2.1-only" instead of "LGPL-2.1-only" for test code
- add SPDX-FileCopyrightText for test code

test/conftest.py
Bastian-Krause: Fixed.

ejoerns (approved):
Ok, that should really be sufficient now regarding licensing :+1:

Issue #78, opened 2021-03-24
GError HTTP Request Failed with Server


MitchGaines:
Hi there,

I have been struggling with the `rauc-hawkbit-updater` service for a few days now and keeping hitting the error below when I try to deploy a new artifact.

Here is my rauc-config.conf file:
```
root@raspberrypi4:~# cat rauc-config.conf
[client]
# host or IP and optional port
hawkbit_server = ***.com
ssl                       = false
ssl_verify                = false
tenant_id                 = DEFAULT
target_name               = dev01 
auth_token                = abc123 #dummy value for actual token 
bundle_download_location  = /tmp/bundle.raucb
retry_wait                = 60
connect_timeout           = 20
timeout                   = 60
log_level                 = debug 
post_update_reboot        = false 

[device]
mac_address               = MACADDRESS
hw_revision               = 1
model                     = dev01
```

Whenever I try to run with the above config file, this is the result:
```
root@raspberrypi4:~# rauc-hawkbit-updater -d -c rauc-config.conf
MESSAGE: Checking for new software...
DEBUG: Response body: {
  "config" : {
    "polling" : {
      "sleep" : "00:05:00"
    }
  },
  "_links" : {
    "deploymentBase" : {
      "href" : "http://***.com/DEFAULT/controller/v1/dev01/deploymentBase/2?c=412523400"
    },
    "configData" : {
      "href" : "http://***.com/DEFAULT/controller/v1/dev01/configData"
    }
  }
}
DEBUG: Identifying ourself to hawkbit server
DEBUG: Request body: {
  "time" : "20210324T125933",
  "status" : {
    "result" : {
      "finished" : "success"
    },
    "execution" : "closed"
  },
  "data" : {
    "model" : "dev01",
    "mac_address" : "MACADDRESS",
    "hw_revision" : "1"
  }
}

(rauc-hawkbit-updater:421): GLib-WARNING **: 12:59:33.452: GError set over the top of a previous GError or uninitialized memory.
This indicates a bug in someone's code. You must ensure an error is NULL before it's set.
The overwriting error message was: HTTP request failed: No error
DEBUG: Failed to get resource from hawkbit server. Status: 400
CRITICAL: Error: HTTP request failed: No error
```

[...]

PR #84, opened 2021-04-14
config-file: use libeconf for config parsing, allow loading default locations

Bastian-Krause:
Pros:
- allows loading config files from default locations like systemd (files in `/etc/` take precedence over those in `/run/` which in turn take precedence over those in `/usr/lib/`) and drop-ins (might be interesting since `auth_token`/`target_name` must be set individually per device)
- trailing spaces as mentioned in #78 are stripped
- parsing booleans is easier

Cons:
- adds a new dependency for [libeconf](https://github.com/openSUSE/libeconf)

Note: this PR is based on #83.


Bastian-Krause:
Another con:
- libeconf is only available in ubuntu >= 20.10 (no LTS with this will be available until April of 2022)

That's why the CI build breaks.

ejoerns:
Rebased onto master (without further changes).

Bastian-Krause:
- dropped test `test_config_trailing_spaces` in favor of existing `test_register_and_check_valid_gateway_token[trailing_space]`
- removed nonexistent `ubuntu-21.04` GitHub actions YAML label (only LTS releases are supported)

Bastian-Krause:
Closing this for now, as suggested above by @ejoerns.

PR #89, opened 2021-06-03
Add cancelation support

Bastian-Krause:
This allows to cancel deployments that were not yet received/downloaded/installed. Once the installation has begun, cancelations are rejected.


Bastian-Krause:
Rebased on current master.

PR #90, opened 2021-06-05
docs: reference: document mandatory [device] section, keep syntax consistent

Bastian-Krause:
- mention (non-obvious) mandatory [device] section with at least one key-value pair
- use `option=<type>` syntax consistently

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -116,3 +116,7 @@ Optional options:
 This section allows to set a custom list of key-value pairs that will be used
 as config data target attribute for device registration.
 They can be used for target filtering.
+
+.. important::
+  The [device] section is mandatory and at least one key-value pair must be
+  configured.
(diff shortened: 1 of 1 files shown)


docs/reference.rst
```diff
@@ -116,3 +116,7 @@ Optional options:
 This section allows to set a custom list of key-value pairs that will be used
 as config data target attribute for device registration.
 They can be used for target filtering.
+
+  .. important::
```
ejoerns, line 120: leading whitespace should not be required here as this directive block is at root level, isn't it?

Bastian-Krause: Probably? I don't know. So the "important note" for `post_update_reboot` is wrong, too? That's where I copied from :)

ejoerns: This one is indented intentionally.

Bastian-Krause: Ah right, because the ``post_update_reboot=<boolean>`` functions as some kind of logical section there and the whole paragraph is indented.

ejoerns: couldn't have said it better, yes ;)

Bastian-Krause: Fixed!

ejoerns (approved):

PR #91, opened 2021-06-09
Prepare for release 1.0

prevas-lkmi:
Updated release notes and version tags in sources and scripts.

Added dist package generation in CMake.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,4 @@
-Release 0.1 (in development)
+Release 1.0 (in development)
 ----------------------------
 
 This is the initial release of RAUC hawkBit Updater.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.0.2)
-project( rauc-hawkbit-updater )
+project( rauc-hawkbit-updater VERSION 1.0 )
 
 option(WITH_SYSTEMD "Set to ON to create unit files and enable systemd startup"  OFF)
 option(BUILD_DOC    "Build documentation" OFF)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,8 @@ if (SYSTEMD_FOUND AND "${SYSTEMD_SERVICES_INSTALL_DIR}" STREQUAL "")
 		"${SYSTEMD_SERVICES_INSTALL_DIR}")
 endif ()
 
+add_definitions( -DPROJECT_VERSION=${CMAKE_PROJECT_VERSION} )
+
 pkg_check_modules(CURL REQUIRED libcurl>=7.47.0)
 include_directories(${CURL_INCLUDE_DIRS})
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -22,7 +22,7 @@ copyright = '2021, Lasse Klok Mikkelsen, Enrico Jörns, Bastian Krause'
 author = 'Lasse Klok Mikkelsen, Enrico Jörns, Bastian Krause'
 
 # The full version, including alpha/beta/rc tags
-release = '0'
+release = '1.0'
 
 
 # -- General configuration ---------------------------------------------------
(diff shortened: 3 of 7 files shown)


CMakeLists.txt
```diff
@@ -1,10 +1,26 @@
...
+foreach (line ${GITIGNORE_LINES})
+    if(line MATCHES "^#")
+	    continue()
+    endif()
+    string(REPLACE "." "[.]" line "${line}")
+    string(REPLACE ";" "\\\\;" line "${line}")
+    string(REPLACE "*" ".*" line "${line}")
+    set(GITIGNORE_CONTENTS "${GITIGNORE_CONTENTS};/${line}")
+endforeach ()
+set(CPACK_SOURCE_IGNORE_FILES "/.git/;${GITIGNORE_CONTENTS}")
```
Bastian-Krause, line 19: That's a pattern, so actual dots should be escaped:
```suggestion
set(CPACK_SOURCE_IGNORE_FILES "/\\.git/;${GITIGNORE_CONTENTS}")
```

Maybe also ignore the .gitignore here? Like this:
```suggestion
set(CPACK_SOURCE_IGNORE_FILES "/\\.git/;\\.gitignore;${GITIGNORE_CONTENTS}")
```

Bastian-Krause (changes requested):
Looks good to me otherwise.

CMakeLists.txt
prevas-lkmi: @Bastian-Krause  Good catch. I will leave the .gitignore in the dist package, if not cmake will break when trying to read the .gitignore file.

Bastian-Krause: What about adding `if(EXISTS ..`? Is this too much hassle?

Also, could you replace the space indentation with tabs?

prevas-lkmi: My bad, it is now fixed.

Bastian-Krause: Thanks! Having this, we can add the `.gitignore` to `CPACK_SOURCE_IGNORE_FILES`, right?

ejoerns:
I also came across the `CPack` module for CMake when investigating for packing. What I did not notice is that besides `make package` also a `make package_source` exists. Thus I dropped this idea thinking it would be useful for binary releases only.

But what I do not really get yet, is what the advantage of the current approach is over a `git archive` call?

Having a blacklist for files not to add, will always require having a clean source directory, because each file I do not catch by the blacklist, will make it into the build. And if I have a clean source directory, I do not need any filtering. Am I wrong here?

In rauc with automake for example, we have a distinct whitelist instead. This allows to create and test a distribution, even from non-clean directories.

prevas-lkmi:
@ejoerns there is no advantage and it requires a clean source directory. If you prefer to use git archive, I could remove the last commit?

CHANGES
```diff
@@ -1,3 +1,34 @@
...
+* Add sphinx-based user documentation
+
+.. rubric:: Bug fixes
+
+* Memoryleaks
+* Use-after-frees
+* Raceconditions
+* Fix free storage space calculation overflowing
+
 Release 0.1 (in development)
```
ejoerns, line 32: We could skip this section as it was never really 'released'. Or you tag the initial version as 0.1, but then probably the CHANGES summary for 1.0 is quite unverbose.

ejoerns:
> @ejoerns there is no advantage and it requires a clean source directory. If you prefer to use git archive, I could remove the last commit?

If you ask me, I would leave it out as long as we do not get any notable benefit from it, yes. But final decision is up to you ;)

ejoerns:
So as the testing blockers are resolved now, anything we can do to help for release preparation? :)

Bastian-Krause:
@prevas-lkmi Can I help with anything? Having a first release would be really useful.

ejoerns:
Indeed ;) I'm back from vacation, so I could help, too.

CHANGES
ejoerns: We just skip the changes as this will be the initial official release

ejoerns:
* rebased onto latest master
* updated CHANGES
* removed cpack handling
* generate version information define
* updated some copyright notes

Bastian-Krause (approved):
Looks fine, thanks!

PR #94, opened 2021-06-14
github: use address/leak sanitizer

Bastian-Krause:
Using `-fsanitize=address  -fsanitize=leak -g` can reveal critical issues, so add that to the GitHub Actions run.

`ASAN_OPTIONS=fast_unwind_on_malloc=0` makes sure the stack traces are complete.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,10 +6,13 @@ jobs:
   build:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         options:
         - -DWITH_SYSTEMD=ON
         - -DWITH_SYSTEMD=OFF
+        - -DWITH_SYSTEMD=ON -DCMAKE_C_FLAGS="-fsanitize=address -fsanitize=leak -g" -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address -fsanitize=leak"
+        - -DWITH_SYSTEMD=OFF -DCMAKE_C_FLAGS="-fsanitize=address -fsanitize=leak -g" -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address -fsanitize=leak"
     steps:
     - name: Inspect environment
       run: |
@@ -57,7 +60,7 @@ jobs:
     - name: Run test suite
       run: |
         ./test/wait-for-hawkbit-online
-        dbus-run-session -- pytest -v
+        ASAN_OPTIONS=fast_unwind_on_malloc=0 dbus-run-session -- pytest -v
 
   docs:
     runs-on: ubuntu-latest


Bastian-Krause:
The use-after-free is fixed with #92. The locking issue is fixed with #93.

Bastian-Krause:
Rebased on current master.

A memory leak is remaining. It was there since the beginning, but I did not realize it. I cannot reproduce this locally on Debian 10.9. When I find the time, I'll try to reproduce this in an Ubuntu 20.04 schroot.

ejoerns:
It seems as if we run into a glib upstream leak here. The description sounds quite similar to https://gitlab.gnome.org/GNOME/glib/-/issues/2312

Only builds on machines with newer glib version are affected (e.g. 2.66.8-1 from debian testing).

I would not see this as a release blocker for rauc-hawkbit-updater hence.

Bastian-Krause:
Agreed.

To be able to tell these errors apart, I've updated this PR to run the sanitizer tests in separate matrix jobs.

ejoerns:
I do not see any benefit from merging tests that are not 'green'. No one will have a look at them anyway, but they will probably confuse.
Isn't there a subset that does not cause these failures? Or can we blacklist/skip the failing tests? Then we can make at least sure that no other issues are introduced.

Bastian-Krause:
> I do not see any benefit from merging tests that are not 'green'. No one will have a look at them anyway, but they will probably confuse.

Okay.

> Isn't there a subset that does not cause these failures? Or can we blacklist/skip the failing tests? Then we can make at least sure that no other issues are introduced.

It seems that one or two tests are affected by that. We could probably mark them accordingly. We would need to distinguish the sanitizer run from the non-sanitizer run, however.

The issue you mentioned above suggests to test whether it's "a lack of iteration of the GMainContext before ending the test". This issue seems to occur only in tests where we bail out early (e.g. `test_install_bundle_no_dbus_iface`). That seems to support this theory. I'd still like to try that out.

Bastian-Krause:
> The issue you mentioned above suggests to test whether it's "a lack of iteration of the GMainContext before ending the test". This issue seems to occur only in tests where we bail out early (e.g. `test_install_bundle_no_dbus_iface`). That seems to support this theory. I'd still like to try that out.

This fixes the memory leak:

```diff
diff --git a/src/rauc-installer.c b/src/rauc-installer.c
index 6c042c5..9934614 100644
--- a/src/rauc-installer.c
+++ b/src/rauc-installer.c
@@ -113,6 +113,8 @@ static void install_context_free(struct install_context *context)
 
         g_free(context->bundle);
         g_mutex_clear(&context->status_mutex);
+        // make sure all pending events are processed
+        while (g_main_context_iteration(context->loop_context, FALSE));
         g_main_context_unref(context->loop_context);
         g_assert_cmpint(context->status_result, >=, 0);
         g_assert_true(g_queue_is_empty(&context->status_messages));
```
 Does that look sane to you? Or is this still upstream's fault?

ejoerns:
> > The issue you mentioned above suggests to test whether it's "a lack of iteration of the GMainContext before ending the test". This issue seems to occur only in tests where we bail out early (e.g. `test_install_bundle_no_dbus_iface`). That seems to support this theory. I'd still like to try that out.
> 
> This fixes the memory leaks:
> 
> ```diff
> diff --git a/src/rauc-installer.c b/src/rauc-installer.c
> index 6c042c5..9934614 100644
> --- a/src/rauc-installer.c
> +++ b/src/rauc-installer.c
> @@ -113,6 +113,8 @@ static void install_context_free(struct install_context *context)
>  
>          g_free(context->bundle);
>          g_mutex_clear(&context->status_mutex);
> +        // make sure all pending events are processed
> +        while (g_main_context_iteration(context->loop_context, FALSE));
>          g_main_context_unref(context->loop_context);
>          g_assert_cmpint(context->status_result, >=, 0);
>          g_assert_true(g_queue_is_empty(&context->status_messages));
> ```
> 
> Does that look sane to you? Or is this still upstream's fault?

I had tried this by myself already, but without success. Possible that I placed it wrongly in the code.

If that fixes the errors and as it does not look wrong, I would agree to take it, even if we are not 100% sure who would be actually responsible for handing the remaining events.

Bastian-Krause:
> If that fixes the errors and as it does not look wrong, I would agree to take it, even if we are not 100% sure who would be actually responsible for handing the remaining events.

For the record: the pattern seems to be used [in](https://gitlab.freedesktop.org/gstreamer/gst-plugins-base/-/blob/941629291bb1e2e679f8fbd81b9bc483effdce99/gst-libs/gst/rtsp/gstrtspconnection.c#L2914) [various](https://gitlab.freedesktop.org/gstreamer/gstreamer/-/blob/120e69db60f4ea4501e307c57e04f8fd65fecc3f/tools/gst-launch.c#L1209) [projects](https://review.spdk.io/gerrit/plugins/gitiles/spdk/qemu/+/55a2201e79d063766509ed2f2f2e8837b9e1facf/tests/test-aio.c#484).

Bastian-Krause:
Rebased on top of PR #95.

ejoerns (dismissed):
Since from now on all tests succeed: From my point of view, this PR can be merged/rebased after #95 got merged.

Bastian-Krause:
Rebased on master.

ejoerns (approved):

PR #98, opened 2021-06-30
rauc-hawkbit-updater: support available locales

Bastian-Krause:
This makes it possible to specify host names containing unicode character in the "hawkbit_server" config option if the system supports it.

Otherwise setting "hawkbit_server" to hüseldösel.de fails with (having CURLOPT_VERBOSE enabled):

```
* Failed to convert hüseldösel.de to ACE; could not convert string to UTF-8
* Closing connection -1
WARNING: Scheduled check for new software failed: URL using bad/illegal format or missing URL (3)
```

### DIFF
--- a/src/rauc-hawkbit-updater.c
+++ b/src/rauc-hawkbit-updater.c
@@ -7,6 +7,7 @@
  */
 
 #include <stdio.h>
+#include <locale.h>
 #include <glib.h>
 #include <glib/gprintf.h>
 #include "rauc-installer.h"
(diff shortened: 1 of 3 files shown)


Bastian-Krause:
Force-pushed:
- require locale's encoding to be UTF-8
- add test case for insufficient locale

jluebbe:
Does this add a new requirement to have a UTF-8 locale shipped with the system (they can be quite large)? Or is this required already for other reasons?

Bastian-Krause:
@jluebbe Yes, this introduces a requirement for UTF-8 locales. Do you have a suggestion how to handle this differently?

Bastian-Krause:
We discussed this internally: UTF-8 locales can be quite large. We do not want to enforce that. PR #99 describes how to handle unicode characters in host names instead.

PR #101, opened 2021-08-09
hawkbit-client: download resuming, allow low-speed downloads

Bastian-Krause:
Introduces a new option called `resume_downloads`. If this option is enabled and the bundle download fails with one of the error codes listed as "resumable", the download resumes where it stopped 0.5 seconds later.

Also enable TCP keep-alive probing for bundle downloads. This allow us to determine whether a connection is still up
and running or not. Depending on the outage, this enables us to fail faster. To make this especially useful, introduce the config options `low_speed_time` (similar to `CURLOPT_LOW_SPEED_LIMIT`) and `low_speed_rate` (similar to `CURLOPT_LOW_SPEED_TIME`). If the configuration allows download rates over long time spans, there will be no other intervention for hanging connections until the configured time span is exceeded.

Both efforts combined allow us to download bundles with low transfer speeds and resume them, if necessary (to save traffic).


Bastian-Krause:
Note: based on #89.

Bastian-Krause:
Force-pushed:
- dropped additional check of `hawkbit_config->resume_downloads` in `get_binary()` to simplify resume logic

Bastian-Krause:
Force-pushed changes (most suggested by @ejoerns, thanks!):
- fixed default values for `low_speed_time` and `low_speed_rate` in docs
- dropped check of `fread()` return value prior to calling `g_checksum_update()`
- moved `g_checksum_update()` after `fread()` error check
- moved resume offset detection from `get_binary()` to outer `download_thread()`, `get_binary()` now expects resume offset as argument
- use `g_clear_pointer()` on sha1sum instead of `g_free()`/setting `NULL`
- simplify loop over `resumable_codes`
- extend sleep on resume comment

PR #105, opened 2021-09-15
allow more generic version numbering

ejoerns:
Do not treat version as float which would limit versions to major.minor. Instead handle it as string.

This way it is limited only by the restrictions CMake brings in (format must be major.minor.patch.tweak where each component must be a number).


Bastian-Krause (approved):
Looks good. In the long run, we should revisit #12.

PR #111, opened 2021-10-11
Handle Deployments Maintenance Windows

Bastian-Krause:
If a maintenance window is specifed for a deployment, hawkBit will first signal a new deployment with `deployment.update=skip`. The target is then expected to download the artifact, but skip the actual update for now. When the maintenance window is "available", `deployment.update={soft,forced}` is signalled and the target is expected to update.

To add support for this, store whether to install the update now or later in a new `Artifact` member `do_install` and act accordingly. To keep things simple, follow the usual code flow for delayed bundle installation and `call get_binary()`. It will notice that a bundle exists, do a range request to make sure it is complete and install it subsequently.

Although the deployment's `deployment.download=skip` attribute is currently never passed from hawkBit, it is part of the [DDI API](https://www.eclipse.org/hawkbit/apis/ddi_api/). So implement it by simply skipping the whole deployment when this attribute is received.

### DIFF
--- a/test/helper.py
+++ b/test/helper.py
@@ -73,3 +73,15 @@ def available_port():
         sock.bind(('localhost', 0))
         sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
         return sock.getsockname()[1]
+
+def timezone_offset_utc(date):
+    utc_offset = int(date.astimezone().utcoffset().total_seconds())
+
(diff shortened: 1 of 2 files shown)


Bastian-Krause:
Force-pushed changes (most suggested by @ejoerns, thanks!):
- added optional note to statement parsing `deployment.maintenanceWindow`
- reworked composing of maintenance window messages, made them their content/level consistent

PR #112, opened 2021-10-15
introduce config.h

ejoerns:
Instead of providing defines as compiler arguments, put them in a
config.h file as autotools and meson would do.

Do not include this by each source file individually but pass it as
compiler -include so that it is not accidentally omitted in new files.

Preparation for possible future adaption/extension of the build system.


ejoerns:
This is deprecated with #113 now fully replacing CMake.

PR #113, opened 2021-10-20
Replace CMake by meson

ejoerns:
This is the 2nd attempt to add meson build system support and bases on the work done in #12.

It intends to be a full adaption of what we have in CMake so far and extends the prior approach in various aspects:

* Support installation (of both binary and systemd service file)
* Set 'systemd' as a meson option
* Support for passing release version (in config.h)
* Install sphinx doc
* Apply recent compiler warning settings
* add GitHub Actions test workflow

Currently based on #112 

Note: I've left the various changes compared to #12 as separate commits to ease review. Maybe for the final commit, we can squash it all together as it just newly introduces meson.


ejoerns:
Any comments on this?

meson.build
```diff
@@ -0,0 +1,102 @@
...
+
+doxygen = find_program('doxygen', required : false)
+if get_option('apidoc') and not doxygen.found()
+  error('Need doxygen for building documentation. Install or reconfigure with -Dapidoc=false')
+endif
+
+if doxygen.found()
+  doc_config = configuration_data()
+  doc_config.set('DOXYGEN_OUTPUT', meson.current_build_dir() / 'doxygen')
+  doc_config.set('DOXYGEN_INPUT', meson.source_root())
```
eli-schwartz, line 80: current_source_dir == source_root when running in the root of the project.

But the latter is deprecated in 0.56, see https://mesonbuild.com/Reference-manual_builtin_meson.html#mesonsource_root

project_source_root() is preferred for clarity, but... you need 0.56 for that. current_source_dir works with any version of meson as long as you evaluate it in the right meson.build file.

meson.build
```diff
@@ -0,0 +1,102 @@
...
+
+  doxyfile = configure_file(input : meson.source_root() + '/Doxyfile.in',
```
eli-schwartz, line 82:
```suggestion
  doxyfile = configure_file(input : 'Doxyfile.in',
```

Like any input file, meson looks for it in the same directory as the target is being defined.

meson.build
```diff
@@ -0,0 +1,102 @@
...
+			    output : 'Doxyfile',
+			    configuration : doc_config,
+			    install : false)
+  custom_target('doxygen',
+    output : 'doxygen',
+    input : [doxyfile] + sources_updater,
+    command : [doxygen, doxyfile],
```
eli-schwartz, line 89:
```suggestion
    input : doxyfile,
    command : [doxygen, '@INPUT@'],
    depend_files:  sources_updater,
```

This lets meson reason more clearly about the relationship between the command inputs and the regen dependencies.

docs/meson.build
```diff
@@ -0,0 +1,27 @@
+sphinx = find_program('sphinx-build', required: false)
+
+if get_option('documentation') and not sphinx.found()
+  error('Need sphinx-build for building documentation. Install or reconfigure with -Ddocumentation=false')
```
eli-schwartz, line 4: Why not use a feature option here for find_program?

meson.build
```diff
@@ -0,0 +1,102 @@
+project('rauc-hawkbit-updater', 'c',
+	 version : '1.0',
```
eli-schwartz, line 2: You can add `meson_version: '>=0.50'` here and meson will use an informative error message if people try to build this with too old versions of meson.

Meson will also warn you when you use meson functionality that is newer than 0.50 without bumping the minimum version.

eli-schwartz (commented):
A few things could perhaps be done a bit more efficiently/idiomatically.

ejoerns:
> A few things could perhaps be done a bit more efficiently/idiomatically.

A bit later than I actually intended to, but: thanks for your review!

I've tried to address most of the changes you mentioned.

.github/workflows/tests-meson.yml
```diff
@@ -0,0 +1,91 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        options:
+        - -Dsystemd=enabled
+        - -Dsystemd=disabled
+        flags:
+        - null
+        - CFLAGS="-fsanitize=address -fsanitize=leak -g" LDFLAGS="-fsanitize=address -fsanitize=leak"
```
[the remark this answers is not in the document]
ejoerns: I sticked to `-g` first of all since I would have to rewrite the matrix test otherwise, But I will keep your suggestion in mind!

meson.build
```diff
@@ -0,0 +1,102 @@
...
+  'src/rauc-installer.c',
+  'src/config-file.c',
+  'src/hawkbit-client.c',
+  'src/json-helper.c',
+  'src/log.c',
+]
+
+c_args = '''
+	  -Wformat-nonliteral
+	  -pedantic
```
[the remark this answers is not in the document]
ejoerns: removed my redundant declaration

meson.build
```diff
@@ -0,0 +1,102 @@
...
+'''.split()
+
+foreach arg : c_args
+  add_project_arguments(arg, language: 'c')
+endforeach
+
+systemddep = dependency('systemd', required: get_option('systemd'))
+libsystemddep = dependency('libsystemd', required: get_option('systemd'))
+
+if get_option('systemd').enabled()
```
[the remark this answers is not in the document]
ejoerns: I see. I have changed it as suggested. :+1:

meson.build
```diff
@@ -0,0 +1,102 @@
...
+  configuration : conf
+)
+add_project_arguments('-include'+meson.current_build_dir() / 'config.h', language: 'c')
+
+doxygen = find_program('doxygen', required : false)
+if get_option('apidoc') and not doxygen.found()
+  error('Need doxygen for building documentation. Install or reconfigure with -Dapidoc=false')
+endif
+
+if doxygen.found()
```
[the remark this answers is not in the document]
ejoerns: I have rewritten parts of the doc generation and now also switched to a feature variable here.

meson.build
ejoerns: Switched to `current_source_dir()` as long as we are not on 0.56 as minimal version.

meson.build
ejoerns: Ok, seems to work, too thus I have adapted this as suggested. Thanks!

meson.build
ejoerns: Added as suggested.

ejoerns:
Next I will rebase, squash, and get rid of CMake to change this into a non-RFC PR.

meson.build
ejoerns: removed the needled `source_root()`.

.github/workflows/tests.yml
```diff
@@ -24,15 +25,16 @@ jobs:
...
 
-    - name: Build (with ${{ matrix.options }})
+    - name: Build (with ${{ matrix.options }} ${{ matrix.flags }})
       run: |
-        mkdir -p build
-        cd build
-        cmake ${{ matrix.options }} -DQA_BUILD=ON ..
-        make
-        cd -
+        ${{ matrix.flags }} meson meson_build ${{ matrix.options }} -Dwerror=true
```
Bastian-Krause, line 32: `QA_BUILD` passed more options than `-Werror`. What about them?

.github/workflows/tests.yml
```diff
@@ -41,7 +43,7 @@ jobs:
 
     - name: Install test dependencies
       run: |
-        sudo apt-get install libgirepository1.0-dev nginx-full
+        sudo apt-get install libgirepository1.0-dev python3-virtualenv nginx-full
```
Bastian-Krause, line 46: Seems unrelated and probably wrong?

meson.build
```diff
@@ -0,0 +1,102 @@
+project(
+  'rauc-hawkbit-updater',
+  'c',
+  version : '1.2',
+  meson_version : '>=0.50',
+  default_options: [
+    'warning_level=3',
+    ],
```
Bastian-Krause, line 8: Wrong indentation level.

meson.build
```diff
@@ -0,0 +1,102 @@
...
+  license : 'LGPLv2.1')
```
Bastian-Krause, line 9:
```suggestion
  license : 'LGPLv2.1',
)
```

test/conftest.py
```diff
@@ -22,6 +22,8 @@ def pytest_addoption(parser):
 def env_setup(monkeypatch):
     monkeypatch.setenv('PATH', f'{os.path.dirname(os.path.abspath(__file__))}/../build',
                        prepend=os.pathsep)
```
Bastian-Krause, line 24: Why do we keep this?

Bastian-Krause (changes requested):
Thanks! Feedback below.

There is one more cmake reference in `build-uncrustify.sh` that needs to be updated.

.github/workflows/tests.yml
ejoerns: I've defined `warning_level=3` in the `default_options` which should enable the rest like `-Wextra` `-pedantic`, etc and '-Dwerror` is set explicitly in test

.github/workflows/tests.yml
ejoerns: Good catch. This seems to be a remnant from the rebase and has been removed in af993c71985ca3ff2aeaa32e5d2cf97a8f4f22b8 already.

test/conftest.py
ejoerns: This is a remnant from the version that kept CMake in. Will just use `build/` for meson.

meson.build
ejoerns: Fixed.

meson.build
ejoerns: Ack, fixed.

eli-schwartz:
> A bit later than I actually intended to, but: thanks for your review!

No problem, always happy to offer advice. :)

meson.build
```diff
@@ -0,0 +1,103 @@
+project(
+  'rauc-hawkbit-updater',
+  'c',
+  version : '1.2',
+  meson_version : '>=0.50',
+  default_options: [
+    'warning_level=3',
+  ],
+  license : 'LGPLv2.1',
```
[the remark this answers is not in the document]
ejoerns: Fully agreed. I guess I've added the license info before we switched to SPDX identifiers in the entire project.
I've now made it LGPL-2.1-only as it is set for the rest already.

meson.build
```diff
@@ -0,0 +1,103 @@
...
+  -Wdeclaration-after-statement
+  -Wformat=2
+  -Wshadow
+  -Wno-unused-parameter
+  -Wno-missing-field-initializers
+'''.split()
+
+foreach arg : c_args
+  add_project_arguments(arg, language : 'c')
+endforeach
```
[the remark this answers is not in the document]
ejoerns: Ah, good to know. Somehow I have assumed that this is necessary.

docs/meson.build
```diff
@@ -0,0 +1,23 @@
...
+  'reference.rst',
+  'using.rst'
+]
+
+custom_target('doc',
+  output: 'html',
+  input: doc_src,
+  command: [sphinx, '-b', 'html', meson.current_source_dir(), meson.current_build_dir() / 'html'],
+  build_by_default: get_option('doc').enabled(),
+  install: true,
```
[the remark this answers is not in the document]
ejoerns: I've now made install depend on the same option, too.

My intention was not not always build documentation by default, but do so (and install) when having `-Ddoc=enabled` set explicitly (defaults to `auto`).

meson.build
```diff
@@ -0,0 +1,100 @@
...
+  'src/rauc-hawkbit-updater.c',
+  'src/rauc-installer.c',
+  'src/config-file.c',
+  'src/hawkbit-client.c',
+  'src/json-helper.c',
+  'src/log.c',
+]
+
+c_args = '''
+  -Wformat-nonliteral
```
Bastian-Krause, line 32: This seems to be included in `-Wformat=2` below, see [GCC's warning options](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html).

docs/release-checklist.txt
```diff
@@ -16,14 +16,18 @@ Release
 
     git tag -m 'release v1.0' -s -u 925F79DAA74AF221 v1.0
 
-- create source tar archive::
+- create release tar archive::
 
-    git archive --format=tar.gz -o rauc-hawkbit-updater-1.0.tar.gz --prefix=rauc-hawkbit-updater-1.0/ v1.0
+    mkdir build && cd build
+    meson ..
+    ninja dist
```
Bastian-Krause, line 23:
```suggestion
    meson setup build
    ninja -C build dist
```
I think we should be consistent here and turn this into an explicit `meson setup`, let meson create the build dir und let ninja change into it.

docs/release-checklist.txt
```diff
@@ -16,14 +16,18 @@ Release
...
+
+  The resulting archive will be placed at meson-dist/rauc-hawkbit-updater-<version>.tar.xz
```
Bastian-Krause, line 25:
```suggestion
  The resulting archive will be placed at build/meson-dist/rauc-hawkbit-updater-<version>.tar.xz
```

docs/release-checklist.txt
```diff
@@ -16,14 +16,18 @@ Release
...
 
 - sign (and verify) source archive::
 
-    gpg --detach-sign -u 925F79DAA74AF221 --armor rauc-hawkbit-updater-1.0.tar.gz
-    gpg --verify rauc-hawkbit-updater-1.0.tar.gz.asc
+    gpg --detach-sign -u 925F79DAA74AF221 --armor meson-dist/rauc-hawkbit-updater-<version>.tar.gz
+    gpg --verify meson-dist/rauc-hawkbit-updater-<version>.tar.gz.asc
```
Bastian-Krause, line 30:
```suggestion
    gpg --detach-sign -u 925F79DAA74AF221 --armor build/meson-dist/rauc-hawkbit-updater-<version>.tar.gz
    gpg --verify build/meson-dist/rauc-hawkbit-updater-<version>.tar.gz.asc
```

Bastian-Krause: And you've missed the tar.gz -> tar.xz change in these lines.

meson.build
ejoerns: indeed :+1:

docs/meson.build
```diff
@@ -0,0 +1,23 @@
...
+  subdir_done()
+endif
+
+doc_src = [
+  'conf.py',
+  'changes.rst',
+  'contributing.rst',
+  'index.rst',
+  'reference.rst',
+  'using.rst'
```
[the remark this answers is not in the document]
ejoerns: ok

docs/release-checklist.txt
ejoerns: Agreed.

docs/release-checklist.txt
ejoerns: ack

docs/meson.build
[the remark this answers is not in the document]
eli-schwartz: Yup, makes sense.

Bastian-Krause (approved):
Lookin' good. Once this is merged, we should drop the requirement for the "build (-DWITH_SYSTEMD=OFF)" and "build (-DWITH_SYSTEMD=ON)" checks. I think it's somewhere in the GitHub settings, but I don't have access there.

ejoerns:
> Thanks! Feedback below.
> 
> There is one more cmake reference in `build-uncrustify.sh` that needs to be updated.

This is for building uncrustify. Or which one do you refer to?

ejoerns:
> Lookin' good. Once this is merged, we should drop the requirement for the "build (-DWITH_SYSTEMD=OFF)" and "build (-DWITH_SYSTEMD=ON)" checks. I think it's somewhere in the GitHub settings, but I don't have access there.

Yes.

This is configurable at 'Settings' -> 'Branches' -> 'Branch Protection Rules'

Bastian-Krause:
> > There is one more cmake reference in `build-uncrustify.sh` that needs to be updated.
> 
> This is for building uncrustify. Or which one do you refer to?

For the record: I did not not notice that this does not refer to rauc-hawkbit-updater while skimming through the output of `git grep -i cmake`.

Bastian-Krause:
@eli-schwartz Maybe you have an idea what's causing the meson warning here: https://github.com/rauc/rauc-hawkbit-updater/actions/runs/3235386293/jobs/5299741994#step:6:30 ?

eli-schwartz (approved):
LGTM for meson style, all suggestions I had have been implemented.

eli-schwartz:
> @eli-schwartz Maybe you have an idea what's causing the meson warning here: https://github.com/rauc/rauc-hawkbit-updater/actions/runs/3235386293/jobs/5299741994#step:6:30 ?

Hmm, this only happens during `dist` which seems odd. I think it's fixed by https://github.com/mesonbuild/meson/commit/1420d0daceb10cafb52a7405f157032a5cc811a5 anyway.

ejoerns:
> > @eli-schwartz Maybe you have an idea what's causing the meson warning here: https://github.com/rauc/rauc-hawkbit-updater/actions/runs/3235386293/jobs/5299741994#step:6:30 ?
> 
> Hmm, this only happens during `dist` which seems odd. I think it's fixed by [mesonbuild/meson@1420d0d](https://github.com/mesonbuild/meson/commit/1420d0daceb10cafb52a7405f157032a5cc811a5) anyway.

Ok, thanks for having a look at! Thus this looks like a meson-internal issue and should not be a blocker here :+1:

eli-schwartz:
Lol, for some strange reason it also bisects as fixed by https://github.com/mesonbuild/meson/commit/b25a423a645491e83112929f95c1bd9312458a9a which doesn't seem to have any obvious reason to fix it.

It's ultimately not very meaningful, it just warns you that the option doesn't do anything. So no real harm in purely `meson dist` code for old versions of Meson. I'd definitely ignore it.

PR #115, opened 2021-11-16
hawkbit-client: do not pass NULL format to g_strdup_vprintf() in build_api_url()

Bastian-Krause:
`build_api_url()` can be called with `NULL` to get the base deployment URL. Passing `NULL` as a format argument to `g_strdup_vprintf()` returns `NULL` if `GLIB_USING_SYSTEM_PRINTF` is defined, otherwise it leads to a segmentation
fault. The format argument is explicitly documented as "not nullable", so we rely on undefined behavior here.

Skip over `va_start()`, `g_strdup_vprintf()`, `va_end()` for `path=NULL` to fix that.

### DIFF
--- a/src/hawkbit-client.c
+++ b/src/hawkbit-client.c
@@ -702,9 +702,11 @@ static gchar* build_api_url(const gchar *path, ...)
         g_autofree gchar *buffer = NULL;
         va_list args;
 
-        va_start(args, path);
-        buffer = g_strdup_vprintf(path, args);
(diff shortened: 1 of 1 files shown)


Bastian-Krause:
Force-pushed:
- adjusted commit message with recent information about `GLIB_USING_SYSTEM_PRINTF`.

ejoerns (approved):
Good catch and investigation :+1:

PR #123, opened 2022-03-31
Do not include glibc-specific bits/types/struct_tm.h

zyga:
This fails to compile against musl. The time.h include is sufficient.

Signed-off-by: Zygmunt Krynicki <zygmunt.krynicki@huawei.com>

### DIFF
--- a/src/hawkbit-client.c
+++ b/src/hawkbit-client.c
@@ -26,7 +26,6 @@
 #include <glib/gstdio.h>
 #include <json-glib/json-glib.h>
 #include <libgen.h>
-#include <bits/types/struct_tm.h>
 #include <gio/gio.h>
 #include <sys/reboot.h>
 


Bastian-Krause (approved):
Looks good to me, thanks for contributing!

Having a musl build job for GitHub actions would be nice.

zyga:
I'm not sure how hard that is without pulling in all of Yocto. Let me check and report back.

zyga:
> I'm not sure how hard that is without pulling in all of Yocto. Let me check and report back.

So, on a super-quick look. Assuming we can use actions with a docker image for something recent-ish (like Ubuntu 22.04) we could build against glibc in one run and against musl in another. I have not used actions for a very long time so I'd have to check how to do things in this environment but I'm sure it is doable.

zyga:
> > I'm not sure how hard that is without pulling in all of Yocto. Let me check and report back.
> 
> So, on a super-quick look. Assuming we can use actions with a docker image for something recent-ish (like Ubuntu 22.04) we could build against glibc in one run and against musl in another. I have not used actions for a very long time so I'd have to check how to do things in this environment but I'm sure it is doable.

Musl itself is packaged but we would have to also compile all of the dependencies from source. That may be a bit more complex without doing this via yocto.

Bastian-Krause:
Thanks for looking into this.

> So, on a super-quick look. Assuming we can use actions with a docker image for something recent-ish (like Ubuntu 22.04) we could build against glibc in one run and against musl in another.

I guess Ubuntu 22.04 (expected to be LTS) will be available soon-ish on GitHub actions anyway.

> Musl itself is packaged but we would have to also compile all of the dependencies from source. That may be a bit more complex without doing this via yocto.

I don't have any experience with this. It was just a thought.

Bastian-Krause:
Adding something like this to the [meta-rauc](https://github.com/rauc/meta-rauc) CI (including a recipe for rauc-hawkbit-updater) could also be a possibility.

zyga:
My colleague was doing some level of CI in actions for meta-raspberrypi. Perhaps there's a place where we could hold shared actions code. The problem for him, though, was that the default gratis actions runners are totally unsuitable for the task and he's employing some local resources for CI.

ejoerns (approved):
POSIX time.h manual says.

> The <time.h> header shall declare the tm structure [...]

Thus, ack.

PR #125, opened 2022-04-14
drop virtualenv test dependency, improve README

Bastian-Krause:
Small collection of testing related improvements:
- github/README: drop python virtualenv test dependency
- README: clarify which calls must be run in the venv
- README: mention test dependencies

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -41,7 +41,7 @@ jobs:
 
     - name: Install test dependencies
       run: |
-        sudo apt-get install libgirepository1.0-dev python3-virtualenv nginx-full
+        sudo apt-get install libgirepository1.0-dev nginx-full
         python -m pip install --upgrade pip
         pip install -r test-requirements.txt
 
--- a/README.md
+++ b/README.md
@@ -90,6 +90,7 @@ Test Suite
 Prepare test suite:
 
 ```shell
+$ sudo apt install libgirepository1.0-dev nginx-full
 $ python3 -m venv venv
 $ source venv/bin/activate
 (venv) $ pip install --upgrade pip
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ Test Suite
 Prepare test suite:
 
 ```shell
-virtualenv -p python3 venv
+python3 -m venv venv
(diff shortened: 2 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -41,7 +41,7 @@ jobs:
 
     - name: Install test dependencies
       run: |
-        sudo apt-get install libgirepository1.0-dev python3-virtualenv nginx-full
```
ejoerns, line 44: Was `python3-virtualenv` ever used acutally?

.github/workflows/tests.yml
```diff
@@ -60,7 +60,7 @@ jobs:
     - name: Run test suite
       run: |
         ./test/wait-for-hawkbit-online
-        ASAN_OPTIONS=fast_unwind_on_malloc=0 dbus-run-session -- pytest -v
+        ASAN_OPTIONS=fast_unwind_on_malloc=0 dbus-run-session -- python -m pytest -v
```
ejoerns, line 63: How relates this to venv usage?

.github/workflows/tests.yml
Bastian-Krause: It was described in the README.rst. It is not used in the workflow.

.github/workflows/tests.yml
Bastian-Krause: If anyone has an idea *why* that is, I'm also happy to add that.

Bastian-Krause: > With `virtualenv` a `venv/bin/pytest` exists, but not with `venv`. I should have probably added that to the commit message.

This is wrong. It has nothing to do with `virtualenv` vs. `venv`. Old setuptools versions seem to ignore the `console_scripts` entry point and do not generate `venv/bin/pytest`. I'll drop this change.

Bastian-Krause: Done.

.github/workflows/tests.yml
Bastian-Krause: I've added that bit of info to the commit message.

ejoerns (approved):
Looks fine for me now :+1:

PR #130, opened 2022-04-21
Add HTTP Streaming Support

Bastian-Krause:
RAUC supports HTTP streaming since https://github.com/rauc/rauc/pull/755 (not yet included in a release). Documentation is [available here](https://rauc.readthedocs.io/en/latest/advanced.html#http-streaming). This PR aims to add support for HTTP streaming to rauc-hawkbit-updater.

Add a new config option `http_streaming`. If set, rauc-hawkbit-updater does not download the RAUC bundle, but passes the bundle URL along with the authentication header directly to RAUC. RAUC streams the bundle during installation.

The install/cancel tests have been adjusted to run with and without HTTP streaming.

Note that the default hawkBit configuration limits the number of HTTP range requests to ~1000 per action and 200 per second. The configuration suggestion and the configuration in the GitHub actions workflow have been adjusted to disable these limitations. Even with this, hawkBit generates an "ActionStatus" for each range request. This needs discussion with upstream - this is the reason for marking this PR "RFC" for the time being.

Based on #129.

Note: I think we should wait for a RAUC release which includes HTTP streaming before including this PR in a rauc-hawkbit-updater release.


Bastian-Krause:
Upstream discussion: https://github.com/eclipse/hawkbit/issues/1249

jluebbe:
I've looked mainly at the part related to RAUC, which looks good to me. We won't change the API for streaming in RAUC, so I don't see a reason to wait for a release before merging this (it needs to be enabled explicitly here anyway).

Bastian-Krause:
- rebased
- docs: added notes to other config options when used with `http_streaming=true`

Bastian-Krause:
- added: pass through value of `ssl_verify` config option as `tls-no-verify` argument to `InstallBundle()` D-Bus method for streaming installations
- drop note in reference about `ssl_verify` having no effect when combined with `http_streaming=true`

ejoerns:
Can now be rebased on #129

Bastian-Krause:
- rebased on master

Bastian-Krause:
I received no responses in https://github.com/eclipse/hawkbit/issues/1249. I guess we can review/merge this anyway.

BubuOT:
Any chance you can make a new release with this included, now that rauc 1.8 with incremental updates is out? :-)

ejoerns (commented):
Looks fine for me, too and successfully tested on my target hardware.

:+1: 

I am not fully sure yet if `http_streaming` is a proper config value for this feature because of its scope.

What we intend to say is that *RAUC* handles the download/streaming (instead of rauc-hawkbit-updater). And using HTTP is our scope anyway.

Thus maybe something like `rauc_streaming` or `rauc_bundle_streaming`, or `stream_bundle`?

Or even `download-mode=streaming` vs `download-mode=direct` or similar with the benefit we can require an explicit decision and say what the alternatives are. 

These are just thoughts, if all are happy with `http_streaming` I'd also take it :smirk: 

A minor addition I have is that `bundle_download_location` is still mandatory which quite confusing when using streaming mode. But this could also be made as part of another PR.

I would also like to see this feature more prominent in either the docs or at least in the example in README.md.

Bastian-Krause:
> Looks fine for me, too and successfully tested on my target hardware.

Thanks for reviewing and testing!

> I am not fully sure yet if `http_streaming` is a proper config value for this feature because of its scope.
> 
> What we intend to say is that _RAUC_ handles the download/streaming (instead of rauc-hawkbit-updater). And using HTTP is our scope anyway.
> 
> Thus maybe something like `rauc_streaming` or `rauc_bundle_streaming`, or `stream_bundle`?
> 
> Or even `download-mode=streaming` vs `download-mode=direct` or similar with the benefit we can require an explicit decision and say what the alternatives are.

Yep, naming is hard. I think I like `stream_bundle`, but maybe `download-mode` is even better. I'm not really happy with `direct`, though. Maybe we can come up with another variant for that?

> These are just thoughts, if all are happy with `http_streaming` I'd also take it smirk
> 
> A minor addition I have is that `bundle_download_location` is still mandatory which quite confusing when using streaming mode. But this could also be made as part of another PR.

Correct, I had not thought of this. Let me add a commit for that here.

> I would also like to see this feature more prominent in either the docs or at least in the example in README.md.

I'll see how we can feature the bundle streaming better. Suggestions welcome :)

ejoerns:
> Yep, naming is hard.

Fully agreed.

> I think I like `stream_bundle`, but maybe `download-mode` is even better. I'm not really happy with `direct`, though. Maybe we can come up with another variant for that?

I don't like `direct` either but it was the first thing that came to my mind.

Maybe `download-mode=separate` or `download-mode=temp-location` or `download-mode=store` or `download-mode=conventional` ...

Or we stick to `stream_bundle=true`. More opinions welcome. 

> > These are just thoughts, if all are happy with `http_streaming` I'd also take it smirk
> > A minor addition I have is that `bundle_download_location` is still mandatory which quite confusing when using streaming mode. But this could also be made as part of another PR.
> 
> Correct, I had not thought of this. Let me add a commit for that here.

:+1: 

> > I would also like to see this feature more prominent in either the docs or at least in the example in README.md.
> 
> I'll see how we can feature the bundle streaming better. Suggestions welcome :)

For the `README.md` it might be sufficient to note the key in the example config, like

``` ini
# enable_streaming=true
```

For the documentation I would add a `Streaming Support` chapter above the `Plain Bundle Support` chapter.

It could say something like

```
By default, the rauc-hawkbit-updater will first download the bundle to a temporary
storage location and the invoke RAUC to install the bundle.
To not require reserving extra storage and also potentially save download bandwidth
(together with adaptive updates), the rauc-hawkbit-updater can also leverage RAUC's
built-in HTTP streaming support.

To enable it, set ``enable_streaming=true`` in your rauc-hawkbit-updater configuration.

.. note: rauc-hawkbit-updater will add required authentication headers and options to
   the resulting RAUC D-Bus InstallBundle API call.
```

Bastian-Krause:
Force-pushed:
- rename option `http_streaming` -> `stream_bundle`, adjust variables that are named accordingly
- allow omitting `download_location` when `stream_bundle` is enabled
- add basic test without `download_location` and without `stream_bundle` options
- add docs section about streaming supported as suggested by @ejoerns

src/config-file.c
```diff
@@ -316,6 +319,12 @@ Config* load_config_file(const gchar *config_file, GError **error)
                 return NULL;
         }
 
+        if (!bundle_location_given && !config->stream_bundle) {
+                g_set_error(error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_KEY_NOT_FOUND,
+                            "bundle_download_location is required if stream_bundle is disabled");
```
ejoerns, line 324: To make it more obvious, probably

```suggestion
                            "'bundle_download_location' is required if 'stream_bundle' is disabled");
```

Bastian-Krause: Done.

Bastian-Krause:
Force-pushed:
- Use single quotes in error messages around option names for better comprehensibility (as suggested above)

ejoerns (approved):
Looks fine for me now :rocket:

PR #139, opened 2022-10-13
Prepare 1.3 Release

Bastian-Krause:


CHANGES
```diff
@@ -1,3 +1,15 @@
+Release 1.3 (in development)
+----------------------------------
+
+.. rubric:: Enhancements
+
+* Add bundle streaming support [#130]
```
ejoerns, line 6: Maybe a bit more verbose (we have not so many changes anyway :smirk:)?

```suggestion
* Add option ``enable_streaming=true`` to allow using RAUC's HTTP(S) bundle streaming capabilities instead of downloading and storing bundles separately. [#130]
```

CHANGES
```diff
@@ -1,3 +1,15 @@
+Release 1.3 (in development)
+----------------------------------
```
ejoerns, line 2: does not match title

docs/conf.py
```diff
@@ -18,11 +18,11 @@
 # -- Project information -----------------------------------------------------
 
 project = 'RAUC hawkBit Updater'
-copyright = '2021, Lasse Klok Mikkelsen, Enrico Jörns, Bastian Krause'
+copyright = '2022, Lasse Klok Mikkelsen, Enrico Jörns, Bastian Krause'
```
ejoerns, line 21: Maybe we should use a range here, too?

meson.build
```diff
@@ -1,7 +1,7 @@
 project(
   'rauc-hawkbit-updater',
   'c',
-  version : '1.2',
+  version : '1.3',
```
ejoerns, line 4: So far I've updated this with the release commit, but should be ok to do it here, too.

ejoerns (changes requested):

CHANGES
Bastian-Krause: Will do.

CHANGES
Bastian-Krause: Right, I'll fix it.

docs/conf.py
Bastian-Krause: Will do.

meson.build
Bastian-Krause: I just followed docs/release-checklist.txt.

ejoerns (approved):
Thanks!

PR #160, opened 2023-07-10
Bugfix/download only response

### LINKED ISSUE #159: downloadonly deployment is never finished
When `downloadonly` assignment is pushed from Hawkbit, it's finished with the action status `none`. According to [DDI documentation](https://eclipse.dev/hawkbit/rest-api/rootcontroller-api-guide.html#_post_tenantcontrollerv1controlleriddeploymentbaseactionidfeedback) this would keep action in the state `running` and is never finished (which can be observed in the Hawkbit Management UI). But `downloadonly` should not expect any other activities after download, so we should finish the action after download is complete.

UVV-gh:
Modify unit test and response status for `downloadonly` deployments

Fixes #159

### DIFF
--- a/include/hawkbit-client.h
+++ b/include/hawkbit-client.h
@@ -90,6 +90,7 @@ typedef struct Artifact_ {
         gchar *download_url;          /**< download URL of software bundle file */
         gchar *feedback_url;          /**< URL status feedback should be sent to */
         gchar *sha1;                  /**< sha1 checksum of software bundle file */
+        gchar *maintenance_window;    /**< maintenance flag, possible values: available, unavailable, null */
         gboolean do_install;          /**< whether the installation should be started or not */
 } Artifact;
 
--- a/src/hawkbit-client.c
+++ b/src/hawkbit-client.c
@@ -914,6 +914,17 @@ static gpointer download_thread(gpointer data)
         }
 
         g_mutex_lock(&active_action->mutex);
+        // skip installation if hawkBit asked us to do so
+        if (!artifact->do_install &&
+            (!artifact->maintenance_window || g_strcmp0(artifact->maintenance_window, "available") == 0)) {
+                active_action->state = ACTION_STATE_SUCCESS;
+                if (!feedback(artifact->feedback_url, active_action->id, "File checksum OK.", "success",
+                              "downloaded", &feedback_error))
+                        g_warning("%s", feedback_error->message);
+
+                g_mutex_unlock(&active_action->mutex);
+                return GINT_TO_POINTER(TRUE);
+        }
         if (!feedback_progress(artifact->feedback_url, active_action->id, "File checksum OK.",
                                &error)) {
                 g_warning("%s", error->message);
@@ -1064,6 +1075,7 @@ static gboolean process_deployment(JsonNode *req_root, GError **error)
 
         // handle deployment.maintenanceWindow (only available if maintenance window is defined)
         maintenance_window = json_get_string(resp_root, "$.deployment.maintenanceWindow", NULL);
+        artifact->maintenance_window = g_strdup(maintenance_window);
         maintenance_msg = maintenance_window
                           ? g_strdup_printf(" (maintenance window is '%s')", maintenance_window)
                           : g_strdup("");
@@ -1486,6 +1498,7 @@ void artifact_free(Artifact *artifact)
         g_free(artifact->download_url);
         g_free(artifact->feedback_url);
         g_free(artifact->sha1);
+        g_free(artifact->maintenance_window);
         g_free(artifact);
 }
 
--- a/test/test_download.py
+++ b/test/test_download.py
@@ -137,7 +137,7 @@ def test_download_only(hawkbit, config, assign_bundle):
     assert exitcode == 0
 
     status = hawkbit.get_action_status()
-    assert any(['download' in s['type'] for s in status])
+    assert status[0]['type'] == 'downloaded'
 
     # check last status message
     assert 'File checksum OK.' in status[0]['messages']


src/hawkbit-client.c
```diff
@@ -914,6 +914,17 @@ static gpointer download_thread(gpointer data)
         }
 
         g_mutex_lock(&active_action->mutex);
+        // skip installation if hawkBit asked us to do so
+        if (!artifact->do_install &&
+            (!artifact->do_maintenance || g_strcmp0(artifact->do_maintenance, "available") == 0)) {
```
Bastian-Krause, line 919: How is this related to the maintenance window? Does [downloadonly](https://eclipse.dev/hawkbit/rest-api/targets-api-guide.html#_request_fields_5) mean `deployment.update = skip` and `deployment.maintenanceWindow = available` (or undefined)?

src/hawkbit-client.c
```diff
@@ -914,6 +914,17 @@ static gpointer download_thread(gpointer data)
...
+                active_action->state = ACTION_STATE_SUCCESS;
```
Bastian-Krause, line 920: Does this change anything, really?

src/hawkbit-client.c
```diff
@@ -914,6 +914,17 @@ static gpointer download_thread(gpointer data)
...
+                if (!feedback(artifact->feedback_url, active_action->id, "File checksum OK.", "success",
+                              "downloaded", &feedback_error))
+                        g_warning("%s", feedback_error->message);
```
Bastian-Krause, line 923: Wouldn't this also apply to the "normal code path" in this function? So rather replace the `if !feedback_progress(artifact->feedback_url, active_action->id, "File checksum OK.", &error)) {` below?

src/hawkbit-client.c
```diff
@@ -934,7 +945,6 @@ static gpointer download_thread(gpointer data)
 
                 return GINT_TO_POINTER(TRUE);
         }
-
```
Bastian-Krause, line 937: Seems unrelated.

include/hawkbit-client.h
```diff
@@ -90,6 +90,7 @@ typedef struct Artifact_ {
         gchar *download_url;          /**< download URL of software bundle file */
         gchar *feedback_url;          /**< URL status feedback should be sent to */
         gchar *sha1;                  /**< sha1 checksum of software bundle file */
+        gchar *do_maintenance;        /**< maintenance flag, possible values: available, unavailable, null */
```
Bastian-Krause, line 93: `do_maintenance` sounds like a boolean. Maybe `maintenance_window` is better and closer to the DDI API?

Bastian-Krause (commented):
How do you use `downloadonly`? If you use the downloaded RAUC bundle outside of rauc-hawkbit-updater, how can you be sure that it won't be replaced by another download while you're using it?

src/hawkbit-client.c
UVV-gh: Maintenance is communicated with the same DDI parameters (force + skip) . See the corresponding test case for [download only](https://github.com/eclipse/hawkbit/blob/593a0bb146ff49442c78e786ab969b28b599d81a/hawkbit-rest/hawkbit-ddi-resource/src/test/java/org/eclipse/hawkbit/ddi/rest/resource/DdiDeploymentBaseTest.java#L348) and [maintenance](https://github.com/eclipse/hawkbit/blob/593a0bb146ff49442c78e786ab969b28b599d81a/hawkbit-rest/hawkbit-ddi-resource/src/test/java/org/eclipse/hawkbit/ddi/rest/resource/DdiRootControllerTest.java#L654)
`downloadonly` would have maintenance window `undefined` in this case

include/hawkbit-client.h
UVV-gh: It's actually 3 state like mentioned in the comments, `undefined` does not mean `false`. I'm OK to rename, but wouldn't it be confusing with Target API, which actually defines the window (not just availability).

src/hawkbit-client.c
UVV-gh: Good question. You mean that feedback alone would suffice?

src/hawkbit-client.c
UVV-gh: Normal code path reports `proceeding` status. Or I'm missing the point here?

UVV-gh:
> How do you use `downloadonly`? If you use the downloaded RAUC bundle outside of rauc-hawkbit-updater, how can you be sure that it the download is complete and that it won't be replaced by another download while you're using it?

The backend implementation decides (with the user interaction) when to do the actual update after the bundle is already on the device. The next step is the implementation of `confirmationBase` end-point so a user could confirm the download+install step beforehand.

src/hawkbit-client.c
Bastian-Krause: Ah I see.

src/hawkbit-client.c
Bastian-Krause: Ah, I guess you're right. `success`/`downladed` only applies to the `downloadonly` case.

include/hawkbit-client.h
Bastian-Krause: Hm, the Target API is part of the Management API which we don't implement anyway (apart from `script/hawkbit_mgmt.py`). So I'd opt for renaming it. The DDI API attribute calls it `deployment.maintenanceWindow`, so `maintenance_window` should be okay.

UVV-gh:
> That's not clear to me, how is `downloadonly` involved here?

How would I put the bundle on the device without it? :)
This patch makes sure that download is complete.

But I think I understand your concern. What if another download is started while the bundle is installed outside of rauc-hawkbit-updater? I guess this falls a bit out of scope of this particular fix, but I do have a few thoughts on it:
- Make sure the rauc-hawkbit-updater is stopped when the bundle is installed.
- Enhance rauc-hawkbit-updater to use file name from the URL and not from config file (this might not resolve your original concern though)

Bastian-Krause:
> > That's not clear to me, how is `downloadonly` involved here?
> 
> How would I put the bundle on the device without it? :) This patch makes sure that download is complete.
> 
> But I think I understand your concern. What if another download is started while the bundle is installed outside of rauc-hawkbit-updater?

Exactly.

> I guess this falls a bit out of scope of this particular fix, but I do have a few thoughts on it:
> 
> * Make sure the rauc-hawkbit-updater is stopped when the bundle is installed.

Instead of letting rauc-hawkbit-updater run as a daemon, you could probably call `rauc-hawkbit-updater --run-once` every once in a while and check if a bundle appeared at the download location after it terminated.

> * Enhance rauc-hawkbit-updater to use file name from the URL and not from config file (this might not resolve your original concern though)

src/hawkbit-client.c
UVV-gh: Fixed

include/hawkbit-client.h
UVV-gh: Done

include/hawkbit-client.h
```diff
@@ -90,6 +90,7 @@ typedef struct Artifact_ {
         gchar *download_url;          /**< download URL of software bundle file */
         gchar *feedback_url;          /**< URL status feedback should be sent to */
         gchar *sha1;                  /**< sha1 checksum of software bundle file */
+        gchar *maintenance_window;        /**< maintenance flag, possible values: available, unavailable, null */
```
Bastian-Krause, line 93:
```suggestion
        gchar *maintenance_window;    /**< maintenance flag, possible values: available, unavailable, null */
```
Nit pick..

Bastian-Krause (commented):
Everything else looks good.

include/hawkbit-client.h
UVV-gh: Overlooked after auto-renaming :) Fixed

Bastian-Krause (approved):
Thanks!

PR #163, opened 2023-09-03
test: helper: Print stdout on timeout

UVV-gh:
In case of timeout stdout in github actions is truncated, which makes it more difficult to understand the failure. This patch addresses this issue

### DIFF
--- a/test/helper.py
+++ b/test/helper.py
@@ -66,15 +66,24 @@ def run(command, *, timeout=30):
     logger = logger_from_command(command)
     logger.info('running: %s', command)
 
-    proc = subprocess.run(shlex.split(command), capture_output=True, text=True, check=False,
-                          timeout=timeout)
+    def stdout_print_helper(logger, prefix, stdout):
+        if stdout is None:
+            return
 
-    for line in proc.stdout.splitlines():
-        if line:
-            logger.info('stdout: %s', line)
-    for line in proc.stderr.splitlines():
-        if line:
-            logger.warning('stderr: %s', line)
+        for line in stdout.splitlines():
+            if line:
+                logger.info(f'{prefix}: %s', line)
+
+    try:
+        proc = subprocess.run(shlex.split(command), capture_output=True, text=True, check=False,
+                              timeout=timeout)
+    except subprocess.TimeoutExpired as e:
+        stdout_print_helper(logger, "stdout", e.stdout)
+        stdout_print_helper(logger, "stderr", e.stderr)
+        raise
+
+    stdout_print_helper(logger, "stdout", proc.stdout)
+    stdout_print_helper(logger, "stderr", proc.stderr)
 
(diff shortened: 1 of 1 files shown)


test/helper.py
```diff
@@ -66,8 +66,14 @@ def run(command, *, timeout=30):
...
-    proc = subprocess.run(shlex.split(command), capture_output=True, text=True, check=False,
-                          timeout=timeout)
+    try:
+        proc = subprocess.run(shlex.split(command), capture_output=True, text=True, check=False,
+                              timeout=timeout)
+    except subprocess.TimeoutExpired as e:
+        for line in e.output.splitlines():
+            if line:
+                logger.info('stdout: %s', line)
+        raise
```
ejoerns, line 76: Shouldn't `stderr` then be printed as well?

Might it be an option to assign stoud and stderr to variables and handle printout in a `finally` clause?

UVV-gh: Yeah, printing `stderr` makes sense. I'm not sure I'm getting the point with `finally` block. You mean to handle printout in `finally` _instead_ of `except` block?

ejoerns: I ment combining the output for the good case and the error case, yes.

Something like

``` diff
diff --git a/test/helper.py b/test/helper.py
index b067d6e..232aa5c 100644
--- a/test/helper.py
+++ b/test/helper.py
@@ -66,15 +66,21 @@ def run(command, *, timeout=30):
     logger = logger_from_command(command)
     logger.info('running: %s', command)
 
-    proc = subprocess.run(shlex.split(command), capture_output=True, text=True, check=False,
-                          timeout=timeout)
-
-    for line in proc.stdout.splitlines():
-        if line:
-            logger.info('stdout: %s', line)
-    for line in proc.stderr.splitlines():
-        if line:
-            logger.warning('stderr: %s', line)
+    try:
+        proc = subprocess.run(shlex.split(command), capture_output=True, text=True, check=False,
+                              timeout=timeout)
+        stdout = proc.stdout
+        stderr = proc.stderr
+    except subprocess.TimeoutExpired as e:
+        stdout = e.stdout
+        stderr = e.stderr
+    finally:
+        for line in stdout.splitlines():
+            if line:
+                logger.info('stdout: %s', line)
+        for line in stderr.splitlines():
+            if line:
+                logger.warning('stderr: %s', line)
 
     logger.info('exitcode: %d', proc.returncode)
```
But I have no clue how this actually differs from just printing the error itself and this solution also wouldn't allow to easily re-raise the error.

@Bastian-Krause do you have any better idea maybe?

UVV-gh: Published slightly modified version of your proposal

ejoerns (approved):
Have no idea whether its good python practice or not to define methods in methods, but the approach looks otherwise fine to me :+1:

UVV-gh:
Is there anything else left to fix before merge it?

PR #166, opened 2023-09-27
Add option to authenticate via client certificates updated

flobz:
Hello,
I updated this PR: https://github.com/rauc/rauc-hawkbit-updater/pull/38 to apply @Bastian-Krause recommendation.
If there anything else to before merging ?

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -3,8 +3,10 @@ Using the RAUC hawkbit Updater
 
 .. _authentication-section:
 
+Authentication
+--------------
 Authentication via tokens
--------------------------------
+^^^^^^^^^^^^^^^^^^^^^^^^^
 
 As described on the `hawkBit Authentication page <https://eclipse.dev/hawkbit/concepts/authentication/>`_
 in the "DDI API Authentication Modes" section, a device can be authenticated
@@ -25,8 +27,7 @@ recommended to use this token with care because it can be used to
 authenticate any device.
 
 Authentication via Certificates
--------------------------------
-
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 As can be seen in the system configuration settings of hawkBit, there is a
 third option to authenticate the targets. An "Allow targets to authenticate via
 a certificate authenticated by a reverse proxy" option. To use this
@@ -56,35 +57,38 @@ When the hawkBit server receives the request it checks if:
 - sent fingerprint(s) matches the expected fingerprint(s) which is set
   in the system configuration settings of hawkBit
 
-The client certificate will only be used if no tokens are set and a valid path
+The client certificate will only be used if a valid path
 to a certificate and its key is given in the configuration file.
+You can use token and certficate authentication mutualy, with the certificate being
+used to authenticate to the reverse proxy and the token to authenticate to
+hawkbit.
+
+If the CA of the reverse proxy server is untrusted set ``ssl_verify`` to ``false``.
 
 Here an example of how the configuration file might look like:
 
-[client]
-hawkbit_server            = CN_server_certificate:443
-ssl                       = true
-ssl_verify                = true
-tenant_id                 = DEFAULT
-target_name               = test-target
-auth_token                =
-#gateway_token            = bhVahL1Il1shie2aj2poojeChee6ahShu
-client_cert              = /path/to/client_certificate.pem
-client_key               = /path/to/client_certificate.key
-bundle_download_location  = /tmp/bundle.raucb
-retry_wait                = 60
-connect_timeout           = 20
-timeout                   = 60
-log_level                 = debug
-post_update_reboot        = false
-
-[device]
-product                   = Terminator
-model                     = T-1000
-serialnumber              = 8922673153
-hw_revision               = 2
-key1                      = value
-key2                      = value
+    | [client]
+    | hawkbit_server            = CN_server_certificate:443
+    | ssl                       = true
+    | ssl_verify                = true
+    | tenant_id                 = DEFAULT
+    | target_name               = test-target
+    | client_cert              = /path/to/client_certificate.pem
+    | client_key               = /path/to/client_certificate.key
+    | bundle_download_location  = /tmp/bundle.raucb
+    | retry_wait                = 60
+    | connect_timeout           = 20
+    | timeout                   = 60
+    | log_level                 = debug
+    | post_update_reboot        = false
+    |
+    | [device]
+    | product                   = Terminator
+    | model                     = T-1000
+    | serialnumber              = 8922673153
+    | hw_revision               = 2
+    | key1                      = value
+    | key2                      = value
 
 Streaming Support
 -----------------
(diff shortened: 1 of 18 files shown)


docs/using.rst
```diff
@@ -24,6 +24,68 @@ Although gateway token is very handy during development or testing, it's
...
+
+Here an example of how the configuration file might look like:
+
+[client]
+hawkbit_server            = CN_server_certificate:443
+ssl                       = true
+ssl_verify                = true
+tenant_id                 = DEFAULT
+target_name               = test-target
+auth_token                =
```
Bastian-Krause, line 70: Shouldn't this be commented, too?

src/config-file.c
```diff
@@ -259,12 +261,15 @@ Config* load_config_file(const gchar *config_file, GError **error)
...
+        key_client_cert_exists = get_key_string(ini_file, "client", "client_cert", &config->client_cert, NULL, NULL);
+
+        key_client_key_exists = get_key_string(ini_file, "client", "client_key", &config->client_key, NULL, NULL);
+
+        if (!key_auth_token_exists && !key_gateway_token_exists &&  !(key_client_cert_exists && key_client_key_exists)) {
+                g_set_error(error, 1, 4, "Neither a token nor client certificate are set!");
                 return NULL;
         }
-        if (key_auth_token_exists && key_gateway_token_exists) {
+        else if (key_auth_token_exists && key_gateway_token_exists) {
```
Bastian-Krause, line 272: As far as I can see, this allows various strange combinations of auth/gateway token and cert/key:
- auth_token + client_cert + client_key
- gateway_token + client_cert + client_key

Since there a no later checks, both auth methods would need to be present. Does hawkBit support these cases?

- auth_token + client_cert
- auth_token + client_key
- gateway_token + client_cert
- gateway_token + client_key

These cases lead to auth_token or gateway_token being used, but no warning/error. That should be fixed.

docs/using.rst
```diff
@@ -24,6 +24,68 @@ Although gateway token is very handy during development or testing, it's
...
+
+The client certificate will only be used if no tokens are set and a valid path
+to a certificate and its key is given in the configuration file.
+
+Here an example of how the configuration file might look like:
+
+[client]
+hawkbit_server            = CN_server_certificate:443
+ssl                       = true
+ssl_verify                = true
```
Bastian-Krause, line 67: From https://github.com/rauc/rauc-hawkbit-updater/pull/38#discussion_r651111076:
> What happens if these options are false?

src/hawkbit-client.c
```diff
@@ -256,6 +256,12 @@ static void set_default_curl_opts(CURL *curl)
         curl_easy_setopt(curl, CURLOPT_CONNECTTIMEOUT, hawkbit_config->connect_timeout);
         curl_easy_setopt(curl, CURLOPT_SSL_VERIFYPEER, hawkbit_config->ssl_verify ? 1L : 0L);
         curl_easy_setopt(curl, CURLOPT_SSL_VERIFYHOST, hawkbit_config->ssl_verify ? 1L : 0L);
+
+        if (hawkbit_config->client_cert && hawkbit_config->client_key) {
+                curl_easy_setopt(curl, CURLOPT_SSLKEYTYPE, "PEM");
+                curl_easy_setopt(curl, CURLOPT_SSLCERT, hawkbit_config->client_cert);
+                curl_easy_setopt(curl, CURLOPT_SSLKEY, hawkbit_config->client_key);
```
Bastian-Krause, line 263: The existence of these files was never checked. What happens if they don't exist?

docs/using.rst
```diff
@@ -3,8 +3,8 @@ Using the RAUC hawkbit Updater
 
 .. _authentication-section:
 
-Authentication
---------------
```
Bastian-Krause, line 7: Please leave this section as is and add the sections below as subsections.

Bastian-Krause (changes requested):
In https://github.com/rauc/rauc-hawkbit-updater/pull/38#discussion_r651125623, I said:

> It would be great to have an actual test case for this with a reverse proxy. We already use nginx for some test cases (see test/conftest.py).

This would be really useful.

docs/using.rst
flobz: done

docs/using.rst
flobz: I added the answer here : https://github.com/flobz/rauc-hawkbit-updater/blob/419925f13ac05d311d92e9129661783456726d78/docs/using.rst?plain=1#L66
If the CA of the reverse proxy server is untrusted set ``ssl_verify`` to ``false``.

docs/using.rst
flobz: removed !

src/hawkbit-client.c
flobz: I add a check 118b83d7e6d1d3bec6c91bfb6f9adb75ab106835

src/config-file.c
Bastian-Krause: Okay, this should be documented then. The last 4 scenarios should be fixed.

docs/using.rst
Bastian-Krause: Okay, but what about the `ssl` option? That is required, right? Should be documented and a run time check would also be nice.

src/config-file.c
flobz: Is this enough documented : https://github.com/rauc/rauc-hawkbit-updater/pull/166/files#diff-00d05b0405040981002c129a067f3ae336e345b5b62512322b116f63aec6c341R62 ? 
What do you mean by 'The last 4 scenarios should be fixed." ?

docs/using.rst
flobz: you're right I forgot this check : a059fb3

flobz:
@Bastian-Krause can you restart the "workflow" please ? I would like to check if change made to the pipeline works.

ejoerns:
I've just triggered the CI workflow

flobz:
Ok I fixed the CI but I don't understand why it worked before and why it doesn't work anymore... 
I had to add this https://github.com/rauc/rauc-hawkbit-updater/pull/166/commits/b2d77d92759bdc3d8b1383c01f8ec32c70d68d14. 
Can you restart the workflow please ?

Bastian-Krause:
This will probably need to wait a few more weeks until I have time to review this.

flobz:
> 
> I add test but there is an issue. Hawkbit doesn't support downloading artifacts using http and https in the same time. I add a PR on hawkbit repo [eclipse/hawkbit#1449](https://github.com/eclipse/hawkbit/pull/1449) but for now two hawkbit container config are needed.

It's fixed in hawkbit 3.0.0.

@Bastian-Krause would you be able to do a review soon ?

Bastian-Krause:
@flobz With #169 waiting only for final approval, is there anything in this PR that's left or can we close this?

Bastian-Krause:
I'm assuming there's nothing left in this PR, so I'm closing this. Please re-open if that's incorrect.

PR #169, opened 2024-01-31
Add support for SSL/TLS client authentication

rvdgracht:
Add support for mutual TLS authentication. This is the preferred method of authentication for bosch-iot-suite and the only one that allows you to keep the authenticator in a (f)TPM.

Optionally, an Openssl engine can be configured if required for access to the ssl private key.

### DIFF
--- a/test/rauc_dbus_dummy.py
+++ b/test/rauc_dbus_dummy.py
@@ -103,16 +103,16 @@ class Installer:
         return sha1.hexdigest()
 
     @staticmethod
-    def _get_http_bundle_sha1(url, auth_header):
+    def _get_http_bundle_sha1(url, *, auth_header=None, cert=None, verify=True):
         """Download file from URL using HTTP range requests and compute its sha1 checksum."""
         sha1 = hashlib.sha1()
-        headers = auth_header
+        headers = auth_header or {}
         range_size = 128 * 1024  # default squashfs block size
 
         offset = 0
         while True:
             headers['Range'] = f'bytes={offset}-{offset + range_size - 1}'
-            r = requests.get(url, headers=headers)
+            r = requests.get(url, headers=headers, cert=cert, verify=verify)
             try:
                 r.raise_for_status()
                 sha1.update(r.content)
@@ -130,20 +130,31 @@ class Installer:
         Check that required headers are set, bundle is accessible (HTTP or locally) and its
         checksum matches.
         """
-        if 'http-headers' in args:
+        if 'tls-key' in args and 'tls-cert' in args:
+            cert = (args['tls-cert'], args['tls-key'])
+            if 'tls-no-verify' in args and args['tls-no-verify']:
+                verify = False
+            elif 'tls-ca' in args:
+                verify = args['tls-ca']
+            else:
+                verify = True
+
+            source_sha1 = self._get_http_bundle_sha1(source, cert=cert, verify=verify)
+
+        elif 'http-headers' in args:
             assert len(args['http-headers']) == 1
 
             [auth_header] = args['http-headers']
             key, value = auth_header.split(': ', maxsplit=1)
-            http_bundle_sha1 = self._get_http_bundle_sha1(source, {key: value})
-            assert http_bundle_sha1 == self._get_bundle_sha1(self._bundle)
+            source_sha1 = self._get_http_bundle_sha1(source, auth_header={key: value})
 
             # assume ssl_verify=false is set in test setup
             assert args['tls-no-verify'] is True
-
         else:
-            # check bundle checksum matches expected checksum
-            assert self._get_bundle_sha1(source) == self._get_bundle_sha1(self._bundle)
+            source_sha1 = self._get_bundle_sha1(source)
+
+        # check bundle checksum matches expected checksum
+        assert source_sha1 == self._get_bundle_sha1(self._bundle)
 
     @property
     def Operation(self):
--- a/test/test_mtls.py
+++ b/test/test_mtls.py
@@ -1,0 +1,39 @@
+import pytest
+
+from helper import run
+
+@pytest.mark.parametrize('mode', ('download', 'streaming'))
+def test_install_success_mtls(hawkbit, adjust_config, bundle_assigned, mtls_download_port,
+                              pki_dir, ssl_issuer_hash, rauc_dbus_install_success, mode):
+    """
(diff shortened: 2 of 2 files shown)


rvdgracht:
When using an ssl engine with a slow secure key storage, i.e. OPTEE with pkcs11 TA on a stm32mp151 setting up a TLS connection can take some time (I've seen 5 to 50 seconds). Because rauc-hawkbit-updater creates a new connection to the server for EVERY status update and poll, this can be cumbersome. For that I've opened a different pull request with a change that keeps the connection open between request. See  PR #170

Bastian-Krause:
Did you see #166? My understanding is that this PR is quite similar.

Bastian-Krause (commented):
The changes look good to me, although very similar to #166.

This does not support HTTP streaming installation (`stream_bundle = true`), right? #166 seems to support this, although I don't really see how. I would have expected that the `tls-cert`,` tls-key`, `tls-ca` arguments would be set in RAUC's [InstallBundle()](https://rauc.readthedocs.io/en/latest/reference.html#the-installbundle-method). Would you be willing to add client cert authentication support for HTTP streaming installations?

#166 also contains tests for client cert authentication. It would like to have them included here.

If you don't use the HTTP streaming feature, cherry-picking the test without streaming support would be a way forward. #166 could then be rebased to add HTTP streaming support later.

rvdgracht:
> Did you see #166? My understanding is that this PR is quite similar.
Yes. At the last moment when I was creating the PR.
I decided to create a separate PR anyway since #166 seemed stale, and had no support for pkcs11.

> Would you be willing to add client cert authentication support for HTTP streaming installations?
We're not actually using the streaming installation feature, but I added it anyway.
I added the mtls test as well.

The mtls test is heavily based on the work of @flobz (Florain Bezannier).
Cherry picking his patch wasn't possible, since my API is slightly different and because I wanted to use key/cert supplied through the arguments of InstallBundle() instead of externally through a test fixture.
I took the liberty of chopping Florians patch up in more manageble/reviewable chunks.
I wanted to give Florian credit and added his signed-off to paches where I haven't (or barely) changed his code, I hope thats ok.

flobz:
#166 isn't stale I'm waiting for @Bastian-Krause final review :)

Bastian-Krause:
@rvdgracht Thanks for adding streaming support and testing. Do you want to have a look at the failing tests or should I?

Bastian-Krause:
The overall approach looks good to me, once the test failures and the heap-use-after-free are solved, I can fix up some minor Python formatting and maybe simplify a thing or two. Then, this should be ready.

Bastian-Krause:
I tend to merging this instead of #166: the implementation seems more straight forward and the commits are more comprehensible.

@flobz What do you think of this? Would this work for you, too? Is there anything missing in comparison to #166?

Bastian-Krause:
I have reworked the patches:
- rebased on latest master
- fixed a crustify error
- added a warning for mTLS auth error
- rephrased some sentences in the new mTLS docs section
- simplified `set_auth_curl_ssl()` and `install_loop_thread()`
- added test key material to the repo (instead of re-generating it on each test run)
- dropped a lot of overly complex mTLS test infrastructure

Bastian-Krause (approved):

src/hawkbit-client.c
```diff
@@ -242,6 +242,68 @@ static gboolean set_auth_curl_header(struct curl_slist **headers, GError **error
...
+        return TRUE;
+}
+
+/**
+ * @brief Set Curl options for client authentication
+ *
+ * @param[in] curl Curl handle
+ * @param[out] headers curl_slist** of already set headers
+ * @param[out] error   Error
+ * @return TRUE if authorization method set in config and header was added successfully,
```
ejoerns, line 283: That's a copy of the `set_auth_curl_header()` text. Maybe:

```suggestion
 * @return TRUE if the configured authorization method was set up successfully,
```

src/hawkbit-client.c
```diff
@@ -242,6 +242,68 @@ static gboolean set_auth_curl_header(struct curl_slist **headers, GError **error
...
+ *      TRUE if no authorization method set, FALSE otherwise (error set)
+ */
+static gboolean set_auth_curl(CURL *curl, struct curl_slist **headers, GError **error)
+{
+        gboolean res;
+
+        // Try ssl authentication
+        if (hawkbit_config->ssl_key && hawkbit_config->ssl_cert) {
+                res = set_auth_curl_ssl(curl, error);
```
ejoerns, line 292: On failure, this error will be ignored and potentially overridden.

src/hawkbit-client.c
```diff
@@ -242,6 +242,68 @@ static gboolean set_auth_curl_header(struct curl_slist **headers, GError **error
...
+ * @param[out] error   Error
+ * @return TRUE if authorization method set in config and header was added successfully,
+ *      TRUE if no authorization method set, FALSE otherwise (error set)
+ */
+static gboolean set_auth_curl(CURL *curl, struct curl_slist **headers, GError **error)
+{
+        gboolean res;
+
+        // Try ssl authentication
+        if (hawkbit_config->ssl_key && hawkbit_config->ssl_cert) {
```
ejoerns, line 291: It feels a bit inconsistent for me to have the config checks for `set_auth_curl_ssl()` *before* calling the method and the ones for `set_auth_curl_header()` *as part of* the method.

ejoerns (commented):
Looks good to me, except for the few things I've noted.

Nit: There's a typo in the commit title `test: conftest: add fxtures for a HTTPS nginx proxy` (f*i*xtures).

src/hawkbit-client.c
Bastian-Krause: @rvdgracht @flobz After thinking about this again, why should we fall back to token authentication if `ssl_key`/`ssl_cert` are given? Shouldn't that be an error? I tend to simply return `res` here.

src/hawkbit-client.c
Bastian-Krause: I've added a check for the token auth, too. This should make it more consistent.

src/hawkbit-client.c
Bastian-Krause: Applied.

src/hawkbit-client.c
Bastian-Krause: I've dropped the fallback behavior now. I don't think this should block this PR from getting merged.

PR #170, opened 2024-01-31
Share connection between curl requests

rvdgracht:
Using a curl_share between curl requests allows sharing an open connection and re-use cached DNS, PSL and TLS session id.

This change allows performing multiple requests without having to do re-perform the (full) TLS handshake.

For reference, on a stm32mp151c with OPTEE + pkcs11 TA a full TLS handshake takes ~8 seconds. Mostly due to small pager pool (internal sram) available for OPTEE.

With this change a mTLS curl request take around 60ms after the initial connection has been established.

### DIFF
--- a/src/hawkbit-client.c
+++ b/src/hawkbit-client.c
@@ -70,6 +70,10 @@ static GSourceFunc software_ready_cb;
 static struct HawkbitAction *active_action = NULL;
 static GThread *thread_download = NULL;
 
+static gsize curl_share_initialized = 0;
+static CURLSH *curl_share = NULL;
+static GMutex curl_share_locks[CURL_LOCK_DATA_LAST + 1];
+
 GQuark rhu_hawkbit_client_error_quark(void)
 {
         return g_quark_from_static_string("rhu_hawkbit_client_error_quark");
@@ -85,6 +89,20 @@ GQuark rhu_hawkbit_client_http_error_quark(void)
         return g_quark_from_static_string("rhu_hawkbit_client_http_error_quark");
 }
 
+void curl_share_lock(CURL *handle, curl_lock_data data, curl_lock_access access, void *clientp)
+{
+        g_return_if_fail(data <= CURL_LOCK_DATA_LAST);
+
+        g_mutex_lock(&curl_share_locks[data]);
+}
+
+
+void curl_share_unlock(CURL *handle, curl_lock_data data, curl_lock_access access, void *clientp)
+{
+        g_return_if_fail(data <= CURL_LOCK_DATA_LAST);
+
+        g_mutex_unlock(&curl_share_locks[data]);
+}
 /**
  * @brief Create and initialize an HawkbitAction.
  *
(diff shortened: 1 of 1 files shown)


rvdgracht:
I have some concerns about sharing the connection (CURL_LOCK_DATA_CONNECT) between the mainloop (poller) and the download thread. The documentation states: "It is not supported to share connections between multiple concurrent threads." I did place locking on the shared data including CURL_LOCK_DATA_CONNECT though.
I have succesfully tested this. So the mainloop can poll for status while the download thread is streaming data over the
same connection. Maybe someone with more extensive libcurl knowledge can verify this.

Bastian-Krause:
I think inter-thread connection sharing as it's implemented here should be okay. Maybe we should add `-fsanitize=thread` here, so we stumble upon any issues:

https://github.com/rauc/rauc-hawkbit-updater/blob/76f5a1dad2f42dc778d798bc0e68fcfbce690b65/.github/workflows/tests.yml#L16

rvdgracht:
> Maybe we should add `-fsanitize=thread` here

Added.

Bastian-Krause:
> > Maybe we should add `-fsanitize=thread` here
> 
> Added.

TSAN works by inserting instrumentation into compiled code to detect race conditions. We use glib from Debian in CI, meaning those libraries are not built with TSAN instrumentation, so memory accesses cannot be tracked. This leads to false positives. So suggesting to enable the sanitizer was not a good idea after all.

I'll drop the commit.

Bastian-Krause (dismissed):

Bastian-Krause (approved):

src/hawkbit-client.c
```diff
@@ -256,6 +274,24 @@ static void set_default_curl_opts(CURL *curl)
         curl_easy_setopt(curl, CURLOPT_CONNECTTIMEOUT, hawkbit_config->connect_timeout);
         curl_easy_setopt(curl, CURLOPT_SSL_VERIFYPEER, hawkbit_config->ssl_verify ? 1L : 0L);
         curl_easy_setopt(curl, CURLOPT_SSL_VERIFYHOST, hawkbit_config->ssl_verify ? 1L : 0L);
+
+        if (g_once_init_enter(&curl_share_initialized)) {
+                for (gint i = 0; i <= CURL_LOCK_DATA_LAST; i++)
+                        g_mutex_init(&curl_share_locks[i]);
+
+                curl_share = curl_share_init();
+                g_return_if_fail(curl_share);
```
ejoerns, line 283: We use `g_return_*()` methods for input verification only. Thus better either ignore it (since all potential errors for `curl_share_setopt()` are ignored so far anyway) or use `g_assert_nonnull()`.

src/hawkbit-client.c
```diff
@@ -256,6 +274,24 @@ static void set_default_curl_opts(CURL *curl)
         curl_easy_setopt(curl, CURLOPT_CONNECTTIMEOUT, hawkbit_config->connect_timeout);
         curl_easy_setopt(curl, CURLOPT_SSL_VERIFYPEER, hawkbit_config->ssl_verify ? 1L : 0L);
         curl_easy_setopt(curl, CURLOPT_SSL_VERIFYHOST, hawkbit_config->ssl_verify ? 1L : 0L);
+
+        if (g_once_init_enter(&curl_share_initialized)) {
+                for (gint i = 0; i <= CURL_LOCK_DATA_LAST; i++)
```
ejoerns, line 279: Maybe this deserves a short comment explaining that curl shopt requires a different lock for each data type (that is initialized here)?

src/hawkbit-client.c
```diff
@@ -85,6 +89,20 @@ GQuark rhu_hawkbit_client_http_error_quark(void)
         return g_quark_from_static_string("rhu_hawkbit_client_http_error_quark");
 }
 
+void curl_share_lock(CURL *handle, curl_lock_data data, curl_lock_access access, void *clientp)
+{
+        g_assert_cmpint(data, <=, CURL_LOCK_DATA_LAST);
```
ejoerns, line 94: Here, `g_return_*` might fit better as it's an input parameter validation.

ejoerns (commented):
Looks good to me overall. Just a few notes left about possible enhancements I see.

src/hawkbit-client.c
Bastian-Krause: Done.

ejoerns (approved):

Issue #161, opened 2023-07-26
Reports Success Before Being Verified via Reboot


ashlin4010:
In a typical implementation of RAUC it is the responsibility of the newly updated slots to mark their status good or bad.

rauc-hawkbit-updater to my knowledge does not support reporting success back to hawkbit after the newly updated slot has been validated via a boot or external test, rather it reports success immediately upon a ""successful"" install.

I would consider this a problem as it may falsely label a failed install as successful. This is made worse by Hawkbit's lack of API to update the recorded software version number. So not only is a failed install labeled as successful, Hawkbits is unaware of any rollback the target system has performed and will skip reinstall attempts because it thinks the current software is up to date.
See this other user with similar problems: [https://github.com/eclipse/hawkbit/issues/446](https://github.com/eclipse/hawkbit/issues/446)

With SWUpdate's Hawkbit client, it's the responsibility of the updated OS to report a successful install... I think... or its reporting mechanism is broken.

Am I the only person seeing this problem or am missing something in my implementation?
Is it possible to have rauc-hawkbit-updater only inform hawkbit of a successful update after the update has been truly validated?

ejoerns:
> Am I the only person seeing this problem or am missing something in my implementation?

No, you aren't. I am fully aware that the current approach actually reports the successful installation too early.

> Is it possible to have rauc-hawkbit-updater only inform hawkbit of a successful update after the update has been truly validated?

In principle, this is possible, yes. But to realize this with RAUC in a robust manner, there is a number of puzzle pieces missing.

Parts of what is missing I have documented in the RAUC concept issue https://github.com/rauc/rauc/issues/1114.

The short version is that from RAUC's perspective, we have no existing way yet to mark or detect a specific installation as successful after reboot. The overall concept that would allow such things we referred to as "life-cycle handling".

With the latest v1.10 release of RAUC, an important puzzle piece for your use case already made its way into RAUC: A transaction UUID that allows us to unambiguously identify an installation operation. This could then be used to know which installation to acknowledge.

Thus, as soon as we have all the information in RAUC to tell that we have successfully booted and activated a distinct installation, then it would be straight-forward to wire this into rauc-hawkbit-updater.

flaviens:
Hi!

Following up on this and #1114. We reproduced the early-success behavior end-to-end against a real hawkBit server + the updater, and wanted to flag one thing beyond the single-target case.

[...]

PR #172, opened 2024-03-07
Allow disabling close action after sucessfull update

flobz:
In order to wait for reboot to mark update as succeeded I added an option to disable reporting update as finished successfully.
 
Thanks to that we can use an other program (and maybe one day rauc-hawkbit-updater) at boot to report update result.

##  Related issue: 
Reports Success Before Being Verified via Reboot #161


Bastian-Krause:
Can you elaborate on how the external program determines the appropriate action ID to send update feedback to?

flobz:
Yes, the external program request https://eclipse.dev/hawkbit/rest-api/ddi.html#tag/DDI-Root-Controller/operation/getControllerBase and if there the response contains ["_links"]["deploymentBase"]["href"] then request this link to get the deployment version and the action id to send update feedback to.
I updated the desciprition of the PR to add a chart.

Bastian-Krause:
Sorry for the long delay.

I am okay with the feature. @ejoerns needs to decide whether it's okay to have this as a stop gap solution until we have https://github.com/rauc/rauc/issues/1114 + integration in rauc-hawkbit-updater.

I think the feature only makes sense with the information provided in this PR. So we need to add that in text form to the documentation. Otherwise it might cause more harm than benefit.

After a reboot, rauc-hawkbit-updater must only be started after the external program has terminated successfully to prevent another installation of the same artifact. This needs to be documented.

I'd suggest to wait with feedback integration until @ejoerns commented.

mruettgers:
We'd also appreciate seeing this PR merged, as we're in the same situation - needing to notify Hawkbit of the update state after a reboot and basic system verification.

PR #173, opened 2024-03-14
.readthedocs.yaml: add docs requirements to fix rtd build

Bastian-Krause:
In the past readthedocs installed sphinx-rtd-theme implicitly, but not anymore. So add a requirements.txt with a pinned version of sphinx-rtd-theme and reference it in the .readthedocs.yaml.


jluebbe:
In rauc, we use pip-compile to fix all dependencies:
https://github.com/rauc/rauc/blob/master/docs/requirements.in
https://github.com/rauc/rauc/blob/master/docs/requirements.txt

Otherwise, RTD seems to install versions dependent on the project creation date...

Bastian-Krause:
Even better, I've adopted the same for rauc-hawkbit-updater.

jluebbe (approved):

PR #174, opened 2024-03-15
Allow disabling Authorization headers for bundle downloads

fardke:
Hello,

To be able to download bundle from some external storage (like aws S3 or azure blob storage) it can be useful to not send Authorization header.

In order to test this modification correctly I add a sniffer (using scapy module). It's allow us to check requests send by rauc-hawkbit-updater to hawkbit container and validate headers. To make it works we must add specific capabilities. It's annoying, but I don't find any alternatives.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -152,6 +152,10 @@ Optional options:
 
   Defaults to ``message``.
 
+``disable_download_auth_header=<boolean>``
+  Whether to send Authorization header for download requests.
+  Defaults to ``false``
+
 .. _keyring-section:
 
 **[device] section**
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -47,6 +47,18 @@ Thus rauc-hawkbit-updater can remove these bundles after installation only if
 it they are located in a directory belonging to the user executing
 rauc-hawkbit-updater.
 
+Disable authentication header for downloads
+-------------------------------------------
+
+rauc-hawkbit-updater can download bundle from external storage (hawkbit provide
+the download url). rauc-hawkbit-update Native behavior is to provide
+authentication header like for other requests to DDI hawkbit API. But for some
+external bundle storage (like azure blob storage or aws s3), provide this
+header can cause an error AuthenticationFailed.
+
+To disable it, set ``disable_download_auth_header=true`` in the
+:ref:`sec_ref_config_file`.
+
 systemd Example
 ^^^^^^^^^^^^^^^
 
--- a/include/config-file.h
+++ b/include/config-file.h
@@ -12,24 +12,25 @@
  * @brief struct that contains the Rauc HawkBit configuration.
  */
 typedef struct Config_ {
-        gchar* hawkbit_server;            /**< hawkBit host or IP and port */
-        gboolean ssl;                     /**< use https or http */
-        gboolean ssl_verify;              /**< verify https certificate */
-        gboolean post_update_reboot;      /**< reboot system after successful update */
-        gboolean resume_downloads;        /**< resume downloads or not */
-        gboolean stream_bundle;           /**< streaming installation or not */
-        gchar* auth_token;                /**< hawkBit target security token */
-        gchar* gateway_token;             /**< hawkBit gateway security token */
-        gchar* tenant_id;                 /**< hawkBit tenant id */
-        gchar* controller_id;             /**< hawkBit controller id*/
-        gchar* bundle_download_location;  /**< file to download rauc bundle to */
-        int connect_timeout;              /**< connection timeout */
-        int timeout;                      /**< reply timeout */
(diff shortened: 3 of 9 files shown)


docs/reference.rst
```diff
@@ -152,6 +152,10 @@ Optional options:
 
   Defaults to ``message``.
 
+``disable_download_auth_header=<boolean>``
+  Whether to send Authorization header for download requests.
+  Defaults to ``false``
+
```
Bastian-Krause, line 158: The name of the option is confusing. How about `send_download_auth_header` and let it default to `true`?

docs/using.rst
```diff
@@ -47,6 +47,18 @@ Thus rauc-hawkbit-updater can remove these bundles after installation only if
...
 rauc-hawkbit-updater.
 
+Disable authentication header for downloads
+-------------------------------------------
+
+rauc-hawkbit-updater can download bundle from external storage (hawkbit provide
+the download url). rauc-hawkbit-update Native behavior is to provide
+authentication header like for other requests to DDI hawkbit API. But for some
+external bundle storage (like azure blob storage or aws s3), provide this
+header can cause an error AuthenticationFailed.
```
Bastian-Krause, line 57:
```suggestion
rauc-hawkbit-updater can download bundles from external storage (hawkBit
provides the download URL). rauc-hawkbit-updater's default behavior is to
provide authentication headers like for all other DDI API requests. When using
external bundle storage providers like Azure blob storage or AWS S3, sending of
authorization headers leads to errors and thus needs be disabled.
```

include/config-file.h
```diff
@@ -12,24 +12,25 @@
...
+        gchar* tenant_id;                      /**< hawkBit tenant id */
+        gchar* controller_id;                  /**< hawkBit controller id*/
+        gchar* bundle_download_location;       /**< file to download rauc bundle to */
+        int connect_timeout;                   /**< connection timeout */
+        int timeout;                           /**< reply timeout */
+        int retry_wait;                        /**< wait between retries */
+        int low_speed_time;                    /**< time to be below the speed to trigger low speed abort */
+        int low_speed_rate;                    /**< low speed limit to abort transfer */
+        GLogLevelFlags log_level;              /**< log level */
+        GHashTable* device;                    /**< Additional attributes sent to hawkBit */
```
Bastian-Krause, line 32: No need to touch these lines.

src/config-file.c
```diff
@@ -12,13 +12,14 @@
...
-static const gboolean DEFAULT_SSL_VERIFY  = TRUE;
-static const gboolean DEFAULT_REBOOT      = FALSE;
-static const gchar* DEFAULT_LOG_LEVEL     = "message";
+static const gint DEFAULT_CONNECTTIMEOUT                   = 20;     // 20 sec.
+static const gint DEFAULT_TIMEOUT                          = 60;     // 1 min.
+static const gint DEFAULT_RETRY_WAIT                       = 5 * 60; // 5 min.
+static const gboolean DEFAULT_SSL                          = TRUE;
+static const gboolean DEFAULT_SSL_VERIFY                   = TRUE;
+static const gboolean DEFAULT_REBOOT                       = FALSE;
+static const gchar* DEFAULT_LOG_LEVEL                      = "message";
```
Bastian-Krause, line 21: No need touch these lines.

Bastian-Krause, line 21: No need touch these lines.

Bastian-Krause (changes requested):
Sorry for the long waiting time.

The feature itself looks good. Please see my comments below.

I don't really like the scapy testing approach. We could add an nginx include instead that makes nginx respond with an error if auth headers are detected, mimicking what the problematic storage providers do. If you'd like to add that, please base your work on #188.

docs/using.rst
fardke: done

include/config-file.h
fardke: done

src/config-file.c
fardke: done

fardke:
> Sorry for the long waiting time.
> 
> The feature itself looks good. Please see my comments below.
> 
> I don't really like the scapy testing approach. We could add an nginx include instead that makes nginx respond with an error if auth headers are detected, mimicking what the problematic storage providers do. If you'd like to add that, please base your work on #188.

Ok I think it's done. Tell me if it's ok for you.

Thank you.

Bastian-Krause:
Now that #169 is merged, setting the new `send_download_auth_header=true` also prevents sending SSL client certificate authentication. Since there are no headers involved, I've renamed the option from `send_download_auth_header` to `send_download_authentication`.

Bastian-Krause (approved):

ejoerns (approved):

Issue #203, opened 2026-01-08
Hawkbit client handles 1 action per poll, even when there are multiple available


1NFR4R3D:
The function `hawkbit_pull_cb` in [hawkbit-client.c](https://github.com/rauc/rauc-hawkbit-updater/blob/2711c0e02781eb9d97285c87197f1bafef75d0d1/src/hawkbit-client.c#L1432) handles only one action (download or cancellation) every time it runs. 

In cases where a device has many pending actions and has a sensible poll interval, the update is delayed for a significantly long time.

For instance, if a device has 6 updates assigned to it, of which 5 are cancelled and 1 is available for download, and it has been configured to poll the server every 15 minutes, it will take one hour and 15 minutes before it starts downloading the update.

While this isn't a problem for any devices in the field and online, it is a problem for devices that are either offline for a long period, or in our case, devices that have not been deployed yet.

I propose a simple patch for this solution for now: simply set the poll period to a much smaller value until no actions remain. In this patch, I have it set to 10 seconds

```
diff --git a/src/hawkbit-client.c b/src/hawkbit-client.c
index 75cbd03..9c27194 100644
--- a/src/hawkbit-client.c
+++ b/src/hawkbit-client.c
@@ -1327,6 +1327,10 @@ static gboolean hawkbit_pull_cb(gpointer user_data)
         g_autoptr(JsonParser) json_response_parser = NULL;
         JsonNode *json_root = NULL;
 
+        gboolean no_downloads = FALSE;
+        gboolean no_cancellations = FALSE;
+
         g_return_val_if_fail(user_data, FALSE);
 
         if (++data->last_run_sec < data->hawkbit_interval_check_sec)
@@ -1379,6 +1383,7 @@ static gboolean hawkbit_pull_cb(gpointer user_data)
                 }
         } else {
                 g_message("No new software.");
+                no_downloads = TRUE;
         }
         if (json_contains(json_root, "$._links.cancelAction")) {
                 res = process_cancel(json_root, &error);
@@ -1386,10 +1391,12 @@ static gboolean hawkbit_pull_cb(gpointer user_data)
                         g_warning("%s", error->message);
                         g_clear_error(&error);
                 }
+        } else {
+                no_cancellations = TRUE;
         }
 
         // get hawkbit sleep time (how often should we check for new software)
-        data->hawkbit_interval_check_sec = json_get_sleeptime(json_root);
[...]

PR #204, opened 2026-01-08
hawkbit-client: poll in 10s if an activity was performed

1NFR4R3D:
The function `hawkbit_pull_cb` in [hawkbit-client.c](https://github.com/rauc/rauc-hawkbit-updater/blob/2711c0e02781eb9d97285c87197f1bafef75d0d1/src/hawkbit-client.c#L1432) handles only one action (download or cancellation) every time it runs. 

In cases where a device has many pending actions and has a sensible poll interval, the update is delayed for a significantly long time.

For instance, if a device has 6 updates assigned to it, of which 5 are cancelled and 1 is available for download, and it has been configured to poll the server every 15 minutes, it will take one hour and 15 minutes before it starts downloading the update.

While this isn't a problem for any devices in the field and online, it is a problem for devices that are either offline for a long period, or in our case, devices that have not been deployed yet.

Duplicate: #203

### DIFF
--- a/include/config-file.h
+++ b/include/config-file.h
@@ -31,6 +31,7 @@ typedef struct Config_ {
         int retry_wait;                   /**< wait between retries */
         int low_speed_time;               /**< time to be below the speed to trigger low speed abort */
         int low_speed_rate;               /**< low speed limit to abort transfer */
+        int hawkbit_interval_active_override;       /**< hawkbit check interval */
         GLogLevelFlags log_level;         /**< log level */
         GHashTable* device;               /**< Additional attributes sent to hawkBit */
         gboolean send_download_authentication; /**< Send security header in download requests */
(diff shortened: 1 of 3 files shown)


1NFR4R3D:
Added the ability to control the override based on a config value instead of hard-coding it in

Bastian-Krause:
Processing multiple actions sounds like a good enhancement.

Instead of adding another config option and polling the base resource again (which is strictly speaking against the DDI API's `$.config.polling.sleep`), can't we process multiple actions in a single polling interval? I'm assuming multiple actions are present in the deploymentBase response, but I haven't verified that.

1NFR4R3D:
Sorry, took a bit longer than expected to get back to this. 

I have only ever seen a single action in `deployemntBase`, so I believe we can't process multiple actions in a single poll.

I understand that this is against `$.config.polling.sleep`, but I'm not sure if there is a better solution than this: it is non-default behaviour to violate the DDI api.

PR #205, opened 2026-01-30
Add identify_on_startup config option

b4u-mw:
When enabled, the client sends device attributes (configData) to the server on startup, regardless of whether the server requests it via the configData link.

This is useful to actively report the current software version (sw_version) after a device reboot.

Default: false (preserves existing behavior)

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ Setup target (device) configuration file:
   #resume_downloads             = false
   #stream_bundle                = false
   #post_update_reboot           = false
+  #identify_on_startup          = false
   #log_level                    = message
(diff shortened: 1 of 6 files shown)


Bastian-Krause:
Sounds useful! Is there any reason why we should make this configurable?

b4u-mw:
It must interact well with the update server and the package building. I had a mismatch between the version reported by the updater and the version known by the update server. This lead to an update loop (updater reported version X, update server sent an update because of the mismatch, update was installed, updater reported the wrong version again and so son) so better to have it configurable.

PR #209, opened 2026-02-25
Add download speed limit configuration option.

benprisby:
For rate-limited environments, it would be convenient to have the ability to cap the download speed for updates at the client side. While most use cases may prefer the client to download as fast as possible, I have ran into cases in embedded systems where it is desirable to impose speed limits on network activity. Being able to do this at an application level rather than interface-wide offers more system configuration flexibility.

This change plumbs through cURL's `CURLOPT_MAX_RECV_SPEED_LARGE` option as a configuration property, where the speed limit is imposed as an average for the transfer. The default is to not impose any speed limit (current behavior).

### DIFF
--- a/include/config-file.h
+++ b/include/config-file.h
@@ -31,6 +31,7 @@ typedef struct Config_ {
         int retry_wait;                   /**< wait between retries */
         int low_speed_time;               /**< time to be below the speed to trigger low speed abort */
         int low_speed_rate;               /**< low speed limit to abort transfer */
+        int download_speed_limit;         /**< average download speed cap for rate limiting */
         GLogLevelFlags log_level;         /**< log level */
         GHashTable* device;               /**< Additional attributes sent to hawkBit */
         gboolean send_download_authentication; /**< Send security header in download requests */
--- a/include/config-file.h
+++ b/include/config-file.h
@@ -31,7 +31,7 @@ typedef struct Config_ {
         int retry_wait;                   /**< wait between retries */
         int low_speed_time;               /**< time to be below the speed to trigger low speed abort */
         int low_speed_rate;               /**< low speed limit to abort transfer */
-        int download_speed_limit;         /**< average download speed cap for rate limiting */
+        gint64 download_speed_limit;      /**< average download speed cap for rate limiting */
         GLogLevelFlags log_level;         /**< log level */
         GHashTable* device;               /**< Additional attributes sent to hawkBit */
         gboolean send_download_authentication; /**< Send security header in download requests */
--- a/src/config-file.c
+++ b/src/config-file.c
@@ -328,6 +328,8 @@ Config* load_config_file(const gchar *config_file, GError **error)
                 return NULL;
         if (!get_key_int(ini_file, "client", "low_speed_time", &config->low_speed_time, 60, error))
                 return NULL;
+        if (!get_key_int(ini_file, "client", "download_speed_limit", &config->download_speed_limit, 0, error))
+                return NULL;
         if (!get_key_bool(ini_file, "client", "resume_downloads", &config->resume_downloads, FALSE,
                           error))
                 return NULL;
--- a/src/config-file.c
+++ b/src/config-file.c
@@ -153,6 +153,46 @@ static gboolean get_key_int(GKeyFile *key_file, const gchar *group, const gchar 
 }
 
 /**
+ * @brief Get 64-bit integer value from key_file for key in group, default_value must be specified,
+ * returned in case key not found in group.
+ *
+ * @param[in]  key_file      GKeyFile to look value up
+ * @param[in]  group         A group name
+ * @param[in]  key           A key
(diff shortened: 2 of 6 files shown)


jluebbe:
We'd prefer to move away from full bundle downloads towards RAUC's native streaming support, as that's a requirement to benefit from adaptive updates. See [`stream_bundle=<boolean>`](https://rauc-hawkbit-updater.readthedocs.io/en/latest/reference.html#:~:text=stream_bundle%3D%3Cboolean%3E) in the reference.

Is there a reason why you need to use the non-streaming case?

benprisby:
@jluebbe Thanks for the response. I don't have any opposition to switching to streaming, though it doesn't seem to be enabled by default. Are there plans to change that in a future release?

For users that still opt for the non-streaming (current default) case, are you opposed to adopting this additional configuration option? If you'd like, I'm happy to explicitly note in the reference that it does not apply when `stream_bundle` is true since you hand the process off to RAUC for that.

Let me know how you'd like to proceed.

jluebbe:
> @jluebbe Thanks for the response. I don't have any opposition to switching to streaming, though it doesn't seem to be enabled by default. Are there plans to change that in a future release?

Changing the default might be too surprising for existing users, but perhaps we should warn if the `stream_bundle` option not set explicitly. That should be a different PR, though.

> For users that still opt for the non-streaming (current default) case, are you opposed to adopting this additional configuration option? If you'd like, I'm happy to explicitly note in the reference that it does not apply when `stream_bundle` is true since you hand the process off to RAUC for that.

Yes, please add a note in the reference and also in the example conf comment.

benprisby:
I've added the note. Likely outside the scope of this PR, but some other options have a similar note in the reference, but not the example configuration file:

- `low_speed_time`
- `low_speed_rate`
- `resume_downloads` (missing from `config.conf.example` entirely)

Just noting for sake of completeness, but not something that needs to be done here.

include/config-file.h
```diff
@@ -31,6 +31,7 @@ typedef struct Config_ {
         int retry_wait;                   /**< wait between retries */
         int low_speed_time;               /**< time to be below the speed to trigger low speed abort */
         int low_speed_rate;               /**< low speed limit to abort transfer */
+        int download_speed_limit;         /**< average download speed cap for rate limiting */
```
Bastian-Krause, line 34: Should be a gint64.

src/config-file.c
```diff
@@ -328,6 +328,8 @@ Config* load_config_file(const gchar *config_file, GError **error)
                 return NULL;
         if (!get_key_int(ini_file, "client", "low_speed_time", &config->low_speed_time, 60, error))
                 return NULL;
+        if (!get_key_int(ini_file, "client", "download_speed_limit", &config->download_speed_limit, 0, error))
```
Bastian-Krause, line 331: `get_key_int()` uses `g_key_file_get_integer()` internally. But it should use `g_key_file_get_int64()` instead, so speeds >= 2 GB/s can be set.

src/hawkbit-client.c
```diff
@@ -403,6 +403,8 @@ static gboolean get_binary(const gchar *download_url, const gchar *file, curl_of
         curl_easy_setopt(curl, CURLOPT_LOW_SPEED_TIME, hawkbit_config->low_speed_time);
         curl_easy_setopt(curl, CURLOPT_LOW_SPEED_LIMIT, hawkbit_config->low_speed_rate);
 
+        curl_easy_setopt(curl, CURLOPT_MAX_RECV_SPEED_LARGE, hawkbit_config->download_speed_limit);
```
Bastian-Krause, line 406: Needs a proper cast to `curl_off_t`.

test/test_download.py
```diff
@@ -193,3 +195,29 @@ def test_download_only_without_auth_header(hawkbit, adjust_config, assign_bundle
...
+    bundle_size = os.path.getsize(rauc_bundle)
+    speed_limit = 100 * 1024  # 100KB/s
+    min_time = bundle_size / speed_limit * 0.8  # 20% tolerance since cap is across average speed
+
+    config = adjust_config({'client': {'download_speed_limit': str(speed_limit)}})
+    assign_bundle(params={'type': 'downloadonly'})  # Don't attempt installation that would fail
+
+    start_time = time.time()
+    out, err, exitcode = run(f'rauc-hawkbit-updater -c "{config}" -r', timeout=120)
+    elapsed_time = time.time() - start_time
```
Bastian-Krause, line 210: We should use `time.monotonic()` instead of `time.time()`.

Bastian-Krause (changes requested):
Sounds good. Some comments, though. Also, please add your "Signed-off-by".

benprisby:
Thanks for the feedback. I've addressed those comments and cleaned up the earlier commit messages to be more consistent with the rest of the repo.

Bastian-Krause (approved):
Rebased. Thanks for addressing the feedback!

PR #165, opened 2023-09-04
Initial confirmationBase implementation

UVV-gh:
Roast me 😼 

This is a first round of confirmationBase end-point implementation, see more discussions on that in [this thread](https://github.com/rauc/rauc-hawkbit-updater/discussions/156).

The workflow assumes that confirmation can be issued much later that requested, therefore a fixed 5 seconds timeout is used to check whether it was actually issued (in case of a daemon). I believe with more testing a better approach might be implemented. In case of "run once" the main thread simply waits for the confirmation.

What is not there yet, but should be added to improve the functionality:
- Size of an artifact can be sent to a user to provide more info
- [AutoConfirmation](https://eclipse.dev/hawkbit/rest-api/rootcontroller-api-guide.html#_post_tenantcontrollerv1controlleridconfirmationbaseactivateautoconfirm) is not implemented. I guess the DBus interface to a user backend can be extended for that.

Let me know what you think.

PR #217, opened 2026-04-17
hawkbit-client:  Add support for soft rollout action type with optional D-Bus readiness check

a-elkaim:
HawkBit supports four rollout action types: Forced, Soft, Time Forced, and Download Only.

Before this change, the updater did not distinguish between Forced and Soft, all deployments that reached the install step were treated identically regardless of whether hawkBit signalled the update as being at the device's discretion (soft mode).

So I made the following changes to support it:

- Action type detection (`src/hawkbit-client.c`): The updater now infers the action type from the deployment handling fields (download/update) and logs it on each poll cycle. This makes behavior more transparent and is the foundation for type-specific handling.

- Soft update readiness gate: For soft updates (DDI fields download=attempt, update=attempt), the updater can optionally call IsReadyForUpdate() on a configurable D-Bus service (system bus) before proceeding. The rationale is that hawkBit's Soft semantic explicitly leaves the installation decision to the device, so it is appropriate to let another application (such as a UI or workload manager) gate the update (e.g. to defer it until the device is idle or the user has consented).

  - If the service returns True, the update proceeds normally.

  - If it returns False, the update is silently skipped; no error is reported to hawkBit, so the action remains open and the check is retried on the next poll cycle (controlled by retry_wait).

    - Forced and Download Only updates bypass this check entirely — they are applied unconditionally as before.
    - Time Forced actions report as Soft until their deadline; hawkBit then transparently switches them to Forced, so they will bypass the check once forced.

- New configuration keys (`config.conf.example`, `include/config-file.h`, `src/config-file.c`):
  ```
    [client]
    # D-Bus well-known name of the permission service (unset = no check)
    soft_update_check_dbus_service = de.example.SoftUpdateCheck

    # When the service is unreachable: true (default) = force the update, false = skip it
    soft_update_check_force_on_unavailable = true
  ```

- Documentation (README.md): A new Soft Update Permission Check section documents the feature, the expected D-Bus interface, and the required system bus policy file.
  

- Tests (`test/test_soft.py`, `test/soft_update_check_dbus_dummy.py`, `test/conftest.py`): Six integration tests cover the full matrix of scenarios:
  - Soft update with no check configured (unconditional install)
  - Check configured, service grants permission (install proceeds)
  - Check configured, service denies permission (update skipped, hawkBit action left open)
  - Check configured, service unreachable, force_on_unavailable=true (default — install forced through)
  - Check configured, service unreachable, force_on_unavailable=false (update skipped)
  - Forced update with a denying service configured (check bypassed, install proceeds)

A D-Bus dummy service (soft_update_check_dbus_dummy.py) is provided for use in tests and manual verification.

### DIFF
--- a/include/config-file.h
+++ b/include/config-file.h
@@ -34,6 +34,12 @@ typedef struct Config_ {
         GLogLevelFlags log_level;         /**< log level */
         GHashTable* device;               /**< Additional attributes sent to hawkBit */
         gboolean send_download_authentication; /**< Send security header in download requests */
+        gchar* soft_update_check_dbus_service; /**< D-Bus service name to call for soft update
+                                                     permission; NULL disables the check */
+        gboolean soft_update_check_force_on_unavailable; /**< if TRUE (default), proceed with the
+                                                               update when the D-Bus service is
+                                                               unreachable; if FALSE,
+                                                               skip the update */
 } Config;
 
 /**
--- a/include/config-file.h
+++ b/include/config-file.h
@@ -40,6 +40,11 @@ typedef struct Config_ {
                                                                update when the D-Bus service is
                                                                unreachable; if FALSE,
                                                                skip the update */
+        gint soft_update_check_unavailable_max_retries;  /**< when force_on_unavailable is FALSE,
+                                                               number of consecutive poll cycles
+                                                               with an unreachable service after
+                                                               which the update is forced anyway;
+                                                               0 (default) disables forced retry */
 } Config;
 
 /**
--- a/README.md
+++ b/README.md
@@ -116,6 +116,8 @@ soft_update_check_force_on_unavailable = true
 
 If `soft_update_check_dbus_service` is not set (the default), soft updates are applied unconditionally.
 
+When `soft_update_check_force_on_unavailable` is `false` and `soft_update_check_unavailable_max_retries` is set to a positive integer, the updater will force the update after that many consecutive poll cycles where the D-Bus service was unreachable. This prevents a permanently unavailable service from blocking updates indefinitely. The retry counter resets when the service becomes reachable again or when a new deployment is received.
+
 **D-Bus policy file (system bus):**
 
 Because the updater connects to the **system bus**, the D-Bus daemon will reject calls to the
--- a/README.md
+++ b/README.md
@@ -49,6 +49,8 @@ Setup target (device) configuration file:
   #post_update_reboot           = false
   #log_level                    = message
   #send_download_authentication = true
+  #soft_update_check_dbus_service = de.example.SoftUpdateCheck
+  #soft_update_check_force_on_unavailable = true
 
   [device]
   product                   = Terminator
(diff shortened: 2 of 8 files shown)


Bastian-Krause:
Just skimming through this: I think this has some overlaps with #165, which implements the DBus interface on rauc-hawkbit-updater's side. Due to the size and limited time, we have not had the chance to review that unfortunately. But your approach seems interesting. Maybe @UVV-gh has an opinion on this one?

I guess you'll need some patience until this gets reviewed, #165 has priority because it's open for so long. Ideally, we find a common, consistent DBus concept for both of them until they are merged.

a-elkaim:
> Just skimming through this: I think this has some overlaps with #165, which implements the DBus interface on rauc-hawkbit-updater's side. Due to the size and limited time, we have not had the chance to review that unfortunately. But your approach seems interesting. Maybe @UVV-gh has an opinion on this one?
> 
> I guess you'll need some patience until this gets reviewed, #165 has priority because it's open for so long. Ideally, we find a common, consistent DBus concept for both of them until they are merged.

When looking at #165 I figured it was another feature, it wasn't clear to me if this was related to the soft/forced OTA types.

The end goal seems similar, so maybe we can indeed find a common, consistent DBus concept for both of them.

In the meantime, I am using my own branch to avoid having soft updates applied while the system is in use (I am working on a sort of video player that is using a schedule to play pre-defined content, if a soft OTA arrives, it should be applied outside of the active schedule).

Let me know if I can assist in further developments.

include/config-file.h
```diff
@@ -34,6 +34,17 @@ typedef struct Config_ {
         GLogLevelFlags log_level;         /**< log level */
         GHashTable* device;               /**< Additional attributes sent to hawkBit */
         gboolean send_download_authentication; /**< Send security header in download requests */
+        gchar* soft_update_check_dbus_service; /**< D-Bus service name to call for soft update
+                                                     permission; NULL disables the check */
+        gboolean soft_update_check_force_on_unavailable; /**< if TRUE (default), proceed with the
```
UVV-gh, line 39: I would actually prefer a safer default here, meaning **false**. And it looks like your commit message claims the opposite, or I'm misreading something.

UVV-gh:
Yeah, it's ultimately the same goal just implemented differently. When I worked on it and [discussed it with the maintainers](https://github.com/eclipse-hawkbit/hawkbit/issues/1039) I didn't get any semantics on the soft assignment behavior, therefore we opted for download-only path and a service that could communicate separately to the server when the update is installed. 

> The rationale is that hawkBit's Soft semantic explicitly leaves the installation decision to the device, so it is appropriate to let another application (such as a UI or workload manager) gate the update (e.g. to defer it until the device is idle or the user has consented).

Are you sure that's what Hawkbit expects? ;) The thing is it's hardly documented anywhere expect the actual source code state machine. What do you want to return to Hawkbit in https://hawkbit.eclipse.dev/rest-api/ddi.html#tag/DDI-Root-Controller/operation/postDeploymentBaseActionFeedback ?
I guess we decided to use download-only approach because there's a possibility in Hawkbit to [switch from soft to force](https://hawkbit.eclipse.dev/rest-api/mgmt.html#tag/Targets/operation/updateAction) and that is unacceptable in our case. And later they introduced ConfirmationBase endpoint, which was enough for us.

@Bastian-Krause I'd say this PR complements mine, because "soft" rollouts are present in both, DeploymentBase and ConfirmationBase APIs. I'm still not sure what "soft" would mean in case of ConfirmationBase :)

include/config-file.h
a-elkaim: Good catch ! My first commit message was containing an error regarding the default value, it is now fixed :)

That being said, the logic I had in mind is the following:

- If `soft_update_check_force_on_unavailable` is set to `true`, then a soft update could still be applied immediately upon failing to reach the service. The idea is not to block a potential critical update (one could argue that a critical update should be forced and not soft).

- Else, if set to `false`, we can rely on `soft_update_check_unavailable_max_retries` (if set) to still apply an update after N consecutive failures.

What do you think ?

